Issue #401, opened 2026-04-11
Codify variety-tiered blocking teachback using variety_scorer.py as backbone


michael-wojcik:
## Context

PACT's current teachback protocol requires dispatched specialists to send a teachback (a restatement of their understanding of the assigned task) before beginning implementation. In practice, the protocol is **advisory, not blocking**: the teammate sends the teachback and proceeds immediately, without waiting for the orchestrator to approve or correct.

This creates a well-documented failure mode: **rote-ritual teachback**. When there is no consequence for sending a shallow teachback, and no consequence for proceeding without orchestrator feedback, teammates treat the teachback as a compliance checkbox. The orchestrator is supposed to validate teachbacks against the dispatch prompt, but the validation window is short (the teammate is already working) and orchestrators routinely skip the validation entirely. The result: misunderstandings disguised as agreement go undetected until TEST phase or peer review, at which point the wasted work is significant.

The [existing handoff_gate hook](../hooks/handoff_gate.py) demonstrates that **mechanical enforcement** of a gate is effective — teammates cannot bypass it by intent or accident. A similar mechanism for teachback would eliminate the discretion gap at its source.

### What exists

- **`pact-plugin/hooks/shared/variety_scorer.py`** — deterministic variety scoring module with canonical thresholds:
  - `validate_dimension(value, name)` — enforces int ∈ [1, 4]
  - `score_variety(novelty, scope, uncertainty, risk) -> int` — returns sum ∈ [4, 16]
  - `route_workflow(score) -> str` — maps score to `comPACT` / `orchestrate` / `plan-mode` / `research-spike`
  - `LEARNING_II_MIN_MATCHES = 5` and `LEARNING_II_MAX_BUMP = 1` — Learning II parameters baked in

- **Feature-level variety scoring** — formalized in `/PACT:orchestrate`, `/PACT:comPACT`, `/PACT:plan-mode` command entry points. The feature task can carry `metadata.variety_score` (convention, not enforced).

- **Learning II calibration** — secretary gathers calibration metrics during HANDOFF processing, records per-domain drift in `pact-memory`. Scoring accuracy improves over time via deutero-learning feedback loop.

- **`handoff_gate` hook** — mechanical enforcement precedent. Teammates cannot escape HANDOFF requirements by ignoring them; the hook intercepts at the tool layer.

### What's missing

[...]

PR #407, opened 2026-04-13
feat(#401): add bootstrap gate enforcement

michael-wojcik:
## Summary

- Add two new hooks enforcing PACT bootstrap as the orchestrator's first action in every session
- `bootstrap_prompt_gate.py` (UserPromptSubmit): injects bootstrap instruction alongside each user message until bootstrap-complete marker exists
- `bootstrap_gate.py` (PreToolUse): blocks code-editing tools (Edit, Write) and agent spawning (Agent) until bootstrap marker exists; exploration tools (Read, Glob, Grep, Bash) remain available
- Marker-write Bash command in `bootstrap.md` self-disables both gates after bootstrap completes
- `session_init.py` clears marker on compaction/context-clear to re-engage gates
- Strengthened CLAUDE.md routing block references the mechanical gate (Layer 4 defense-in-depth)
- 77 new tests (CRITICAL tier, 90%+ coverage): fail-open 100%, non-PACT passthrough 100%, gate logic 90%+

## Four-Layer Architecture

| Layer | Mechanism | File |
|-------|-----------|------|
| 1 | Shell alias (synthetic first message) | User setup |
| 2 | UserPromptSubmit injection | `bootstrap_prompt_gate.py` |
| 3 | PreToolUse blocking gate | `bootstrap_gate.py` |
| 4 | Strengthened routing block | `claude_md_manager.py` |

## Key Design Decisions

- **Bash is NOT blocked**: The marker-write mechanism is a Bash command in bootstrap.md — blocking Bash would create a circular dependency where the gate can never self-disable
- **No PostToolUse marker writer**: PREPARE research confirmed Skill does not fire PreToolUse/PostToolUse (only 11 built-in tools are hookable). Marker write moved to Bash command in bootstrap.md
- **Matcherless PreToolUse registration**: bootstrap_gate.py fires for ALL hookable tools; gate logic handles classification internally

## Test plan

- [x] All 6130 tests pass (77 new + 6053 existing)
- [ ] Verify bootstrap_prompt_gate injects instruction when no marker
- [ ] Verify bootstrap_gate blocks Edit/Write/Agent when no marker
- [ ] Verify both hooks are no-ops for non-PACT sessions
- [ ] Verify both hooks are no-ops for teammate sessions
- [ ] Verify marker cleanup on compaction re-engages gates
- [ ] Verify fail-open on all exception paths


michael-wojcik:
## Historical note: PR mislabeling

This PR's commit messages and body reference \`#401\` and \"Closes #401\", but that's incorrect:

- **Issue #401** is about the **teachback gate** (blocks task completion until teachback is sent, parallel to \`handoff_gate.py\`) — a different feature than what this PR implemented
- **This PR** shipped the **bootstrap gate** (blocks tool calls until \`Skill(\"PACT:bootstrap\")\` is invoked)

The \`(#401)\` tag came from an incorrect association in pact-memory from a prior plan-mode consultation session that generated \`docs/plans/bootstrap-gate-enforcement-plan.md\` without creating a dedicated issue. When this session resumed, the secretary's memory briefing told the orchestrator \"issue #401 = bootstrap gate,\" which was not verified against the actual issue body until wrap-up.

**Corrected tracking:**
- **Issue #401** has been reopened to continue tracking the teachback gate work
- **Issue #409** was created as an after-the-fact tracker for this bootstrap gate feature (closed as IMPLEMENTED)
- **Issue #408** tracks the follow-up stuck-gate detection observability work

The commits and merge can't be rewritten on \`main\`, but this comment is here so future archaeologists understand the labeling history.

Issue #751, opened 2026-05-13
scan-pending-tasks: warmup-grace skip during the first ~120s after arm


michael-wojcik:
## Problem

`/PACT:start-pending-scan` registers a cron with shape `*/2 * * * *` — fires at wall-clock minute marks 0, 2, 4… The first fire after arming lands anywhere from 0 to ~120 seconds after `CronCreate`, depending on when the arm crosses a minute boundary. Empirically (this session: pact-126c99e9, 2026-05-13), the first fire usually lands within seconds of arming, while teammates are still claiming and reading their tasks — they have not yet had time to submit teachback or HANDOFF metadata. The first scan fire is structurally guaranteed to be empty: it's pure overhead.

At the documented 2-minute cadence, 30 fires/hour during active teammate work, an immediate-after-arm fire is one extra empty-scan turn for every dispatch boundary. The Emit-Nothing-If-Empty guardrail keeps the transcript silent, but the LLM turn cost is paid every time (per the §Known Limitations Empty-scan turn cost trade-off).

## Proposed fix (Option 1 — warmup-grace check inside scan body)

Persist an `armed_at` epoch timestamp at arm-time and have the scan body skip-emit-nothing while `now - armed_at < 120s`. Pure addition to existing guardrails; no cron-shape change.

### Sketch

**`/PACT:start-pending-scan` (commands/start-pending-scan.md)**

After the `CronCreate(...)` call (or as part of the same atomic cold-start block):

```python
import json, time, os
from pathlib import Path

session_dir = Path(os.environ.get("CLAUDE_SESSION_DIR", "/tmp"))  # or read from pact-session-context.json
state_file = session_dir / "pending-scan-armed-at.json"
state_file.write_text(json.dumps({"armed_at": time.time()}))
```

Idempotent: when reusing an existing armed cron, leave `armed_at` unchanged (don't overwrite — the original arm time is what the warmup grace measures against). Specifically: check for `state_file` existence BEFORE writing; only write on cold-start (when `CronList` filter shows no `/PACT:scan-pending-tasks` entry).

**`/PACT:scan-pending-tasks` (commands/scan-pending-tasks.md)**

Add Step 0 (before TaskList enumeration):

```python
import json, time
from pathlib import Path

state_file = Path(os.environ.get("CLAUDE_SESSION_DIR")) / "pending-scan-armed-at.json"
try:
    armed_at = json.loads(state_file.read_text())["armed_at"]
[...]

Issue #437, opened 2026-04-15
Test coverage additions for reaper bundle


michael-wojcik:
Follow-up from PR #433 blind-round review cluster. Test-coverage gaps on non-blocking behaviors.

## Items

**Test G2 — `_dir_max_child_mtime` fallback-also-raises path**
Current tests exercise: (a) happy path with readable children, (b) empty dir → parent fallback. NOT exercised: children present BUT every `child.lstat()` raises, AND parent stat also raises. Docstring says this re-raises OSError to caller's per-entry try/except → caller increments `skipped`. Test: mock both child and parent stat to raise, assert `skipped += 1` on caller side. (Note: cycle-5 adds a sentinel for the all-child-raised case; G2 is specifically about the both-child-AND-parent-raised case.)

**Test G4 — integration test with production default base paths**
All current tests override `teams_base_dir`/`tasks_base_dir` with `tmp_path`. A wiring bug in `Path.home() / ".claude" / "teams"` default construction would not be caught. Low risk because construction mirrors `cleanup_old_sessions`, but an integration test pinning the default paths (via `monkeypatch.setattr(Path, "home", lambda: tmp_path)`) would close the gap.

**Test G5 — `reaped` counter semantic**
Current docstring says `reaped` counts *attempted* deletions (counter increments after `rmtree(..., ignore_errors=True)` regardless of success). Not directly tested. Test: mock `rmtree` to silently fail (simulate EPERM swallowed by ignore_errors=True) while `entry.exists()` still returns True after. Assert `reaped` still incremented, dir still on disk. Pins the documented semantic against a future refactor that might switch to verified-deletion counting.

## Priority

Low. All three are coverage completeness improvements; no known regression they'd catch today.

PR #752, opened 2026-05-13
fix(hooks/shared): exclude unowned, orphan, and lead-owned tasks from wake tally (#743)

### LINKED ISSUE #743: hook: _lifecycle_relevant counts lead-owned/unowned umbrella tasks, preventing 1->0 teardown of pending-scan
## Problem

The `/PACT:stop-pending-scan` teardown doesn't fire consistently when expected. During a `/PACT:comPACT` workflow or the CODE phase of `/PACT:orchestrate`, the feature-level task stays `in_progress` even after all dispatched teammate tasks complete (because peer-review hasn't started yet). Similarly during `/PACT:peer-review`, the overarching "review PR" task stays open until merge. In both cases, the cron-based pending-scan continues firing every 2 minutes — accumulating empty-scan turns until session end — instead of being torn down at the natural 1→0 last-active-teammate-task transition.

### Root cause

`pact-plugin/hooks/shared/wake_lifecycle.py` `_lifecycle_relevant` counts ANY task in `pending`/`in_progress` status, with carve-outs only for:

- Wake-excluded agentTypes (currently EMPTY set — every agentType counts).
- Signal-task pattern (`metadata.completion_type == "signal"` AND `metadata.type in {"blocker", "algedonic"}`).

It does NOT enforce "task owner is a teammate (not the team-lead)." The function's own docstring says "active **teammate** tasks" but the predicate doesn't filter on that.

Concretely, workflow commands create umbrella tasks like:

- `/PACT:comPACT` step 1: feature-level task ("Fix #659: ...") with `status="in_progress"`, no owner. Stays in_progress until the workflow completes (often well past the last teammate task).
- `/PACT:orchestrate` steps 1-2: feature task + phase tasks (PREPARE / ARCHITECT / CODE / TEST). Phase tasks are typically marked complete when phase work completes, but the feature task spans the whole workflow.
- `/PACT:peer-review`: overarching "review PR" task that stays in_progress until merge decision.

All of these stay counted by `count_active_tasks()`, so the predicate-driven 1→0 transition the teardown hook watches for never fires.

### Observable effect

- Cron remains armed long after all teammate tasks complete.
- Lead receives ~30 empty-scan-fire turns per hour for the rest of the workflow (until merge / session end).
- The "scan torn down" assertion in workflow command docs becomes false in practice; teardown only happens via user-typed `/PACT:stop-pending-scan` or session-end.

## Proposed fix (Fix A — leadAgentId discriminator)

In `_lifecycle_relevant`, add a check that the task's `owner` resolves to a member in `~/.claude/teams/{team}/config.json` `members[]` AND that member's `agentId != team_config["leadAgentId"]`.

### Sketch

```python
def _lifecycle_relevant(task: Any, team_name: str = "") -> bool:
    # ...existing status check, wake-excluded agentType check...

    owner = task.get("owner")

    # NEW: owner must be a teammate (in members[]) and NOT the team-lead.
    if not isinstance(owner, str) or not owner:
        return False  # unowned umbrella task — not teammate work
    team_config = _read_team_config(team_name)
    if not team_config:
        return True  # config unreadable — fail-conservative, count
    lead_agent_id = team_config.get("leadAgentId")
    members = team_config.get("members") or []
    matched = next(
        (m for m in members if m.get("name") == owner),
        None,
    )
    if not matched:
        return False  # owner string doesn't match any member — orphan
    if matched.get("agentId") == lead_agent_id:
        return False  # team-lead-owned umbrella task — not teammate work

    # ...existing signal-task carve-out, metadata check...
    return True
```

### Why `leadAgentId`, not `agentType`

Three alternatives considered:

| Discriminator | Robustness | Failure under rePACT sub-orchestrator |
|---|---|---|
| **`leadAgentId`** | ✅ One canonical lead per team config; field already used by `start/stop-pending-scan` §Lead-Session Guard | Holds — sub-orchestrator teammate has `agentType=pact-orchestrator` but is NOT the team's `leadAgentId` |
| `agentType == "pact-orchestrator"` | Simpler one-line check | ❌ Breaks: would exclude a future rePACT sub-orchestrator teammate's tasks (they have `agentType=pact-orchestrator` but are NOT the team-lead) |
| Owner string == "team-lead" | Trivial | ❌ Hardcodes the canonical lead name; brittle |

The `leadAgentId` approach mirrors the discriminator already established at `pact-plugin/commands/start-pending-scan.md` §Lead-Session Guard and `pact-plugin/commands/stop-pending-scan.md` §Lead-Session Guard. Reusing the same single-source-of-truth ("the team config knows who the lead is") avoids two-source-of-truth drift on "what makes the lead the lead."

The rePACT sub-orchestrator case is forward-compatibility insurance grounded in a Claude Code Agent Teams platform constraint: **a team can have at most one team-lead** (the agent that called `TeamCreate`; recorded as `leadAgentId` in the team config). Recursive PACT cycles cannot spawn a nested "team-lead" at the platform layer — they have to dispatch a `pact-orchestrator`-typed *teammate* that runs the inner workflow while remaining structurally subordinate to the outer team-lead.

If a future `/PACT:rePACT` implementation takes this shape, the sub-orchestrator teammate's `agentType` will be `pact-orchestrator` BUT its `agentId` will NOT match the team config's `leadAgentId`. The `leadAgentId` check correctly treats that teammate's tasks as teammate work (counted toward the wake tally), while an `agentType`-based check would silently exclude them and break the 1→0 teardown signal — and worse, the bug would only manifest once rePACT lands and would be hard to attribute back to this predicate. The `leadAgentId` discriminator is invariant under the one-team-lead-per-team platform constraint; the agentType discriminator is not.

### Edge cases (all behaviorally correct)

| Case | Result | Correctness |
|---|---|---|
| Unowned task (orchestrator-created feature/phase/review umbrella) | excluded | ✅ matches intent |
| Team-lead-owned task | excluded via `leadAgentId` match | ✅ matches docstring intent |
| Teammate-owned task | included | ✅ matches docstring intent |
| Owner string doesn't match any member (orphan / stale) | excluded | ✅ probably correct — if teammate is gone, no wake to deliver |
| Team config unreadable | fail-conservative (count) | ✅ matches existing fail-open posture elsewhere in the file (livelock-safety > observability) |

## Acceptance criteria

- [ ] `_lifecycle_relevant` in `pact-plugin/hooks/shared/wake_lifecycle.py` enforces the leadAgentId-based teammate-owner check.
- [ ] Docstring updated to make the contract literally enforced (was "lifecycle-relevant"; now also "teammate-owned, not lead-owned").
- [ ] Unit tests cover all five edge cases in the table above.
- [ ] Behavioral test: simulate the `/PACT:comPACT` umbrella-task scenario (one umbrella unowned task + one teammate task → teammate task completes → `count_active_tasks()` returns 0, not 1).
- [ ] No changes to the Lead-Session Guard logic in start/stop-pending-scan — those already use `leadSessionId` (a different but related field for session-vs-session discrimination, distinct from `leadAgentId` for teammate-vs-lead discrimination).

## Non-goals

- No changes to workflow command bodies (orchestrate, comPACT, peer-review). The fix lives entirely in the hook predicate.
- No introduction of a `metadata.umbrella=true` marker. Fix B (umbrella marker) was considered and rejected — depends on every workflow command remembering to mark umbrella tasks; brittle.

## Cross-references

- Related: `pact-plugin/commands/start-pending-scan.md` §Lead-Session Guard — already uses `leadSessionId` as the canonical discriminator for session-vs-session; this fix reuses the analogous `leadAgentId` for teammate-vs-lead at the task layer.
- Related: `pact-plugin/skills/pact-agent-teams/SKILL.md` — defines teammate vs. team-lead semantics; this fix aligns the hook predicate to that model.
- Forward-compat: a future `/PACT:rePACT` sub-orchestrator dispatch (`pact-orchestrator`-typed teammate leading a nested cycle) would silently break under an `agentType`-based discriminator; the `leadAgentId` approach holds.
- Empirical instance: observed in this session (2026-05-13), confirmed by inspecting `wake_lifecycle.py` `_lifecycle_relevant` against the user's behavioral report.

michael-wojcik:
## Summary

- Adds a teammate-owner filter to `_lifecycle_relevant` in `pact-plugin/hooks/shared/wake_lifecycle.py` so umbrella tasks (orchestrator-created feature/phase/review records, lead-owned tasks, and orphan-owner tasks) no longer count toward the wake-mechanism's active-tally.
- Restores the natural 1→0 last-active-teammate-task transition that `wake_lifecycle_emitter.py` watches for, enabling `/PACT:stop-pending-scan` to fire on schedule instead of accumulating empty-scan turns until session end.
- Closes #743.

## Implementation

Six-commit chain on `fix-743-lifecycle-relevant-leadagentid`:

1. `fc62811b` — Add `_read_team_lead_agent_id` sibling reader in `pact_context.py` (parallels `_iter_members`; silent-on-error returning `\"\"`).
2. `8765d1be` — Predicate fix in `wake_lifecycle.py`: two new private helpers (`_owner_is_team_member`, `_is_lead_owned`) + new step 4 owner-check inserted between the existing wake-excluded-agentType check (step 3) and metadata-shape check (step 5). 3 legacy test assertions inverted in lockstep to match the new orphan-exclusion semantics.
3. `37adadd2` — Dedicated test module covering five edge cases (unowned / lead-owned / teammate-owned / orphan / config-unreadable), a behavioral umbrella scenario (2→1→0 transition), pure-never-raises invariants on both helpers, and the AND-composition pin (name AND agentId both required for lead-ownership). Helpers tightened to guard against non-string `team_name`.
4. `aedf77dd` — Drop two phantom-green structural pins (asserted source-text presence that survived a step-4 revert) and relocate the fail-CONSERVATIVE audit-anchor comment block from before the `if team_name:` body to inside it, so the audit-anchor invariant fires as a true counter-signal under body-only revert.
5. `8c997471` — TEST-phase coverage: adversarial type-coercion on the owner field (stringified non-string-shaped owners, malformed member entries, empty/null `leadAgentId`), 100×20 stress test bounding `_iter_members` calls at 3× per `_lifecycle_relevant` invocation, and a realistic comPACT-shaped integration scenario walking count_active_tasks across the full lifecycle.
6. `3cd6f31e` — Plugin version bump 4.2.2 → 4.2.3 (4-file dance).

## Why `leadAgentId` discriminator (not `agentType`)

Per the issue body and architecture spec:

- One canonical lead per team config (`leadAgentId` is already used by `start/stop-pending-scan` §Lead-Session Guard for analogous discrimination at the session layer).
- Forward-compatible with rePACT sub-orchestrator teammates: a sub-orchestrator's `agentType` would be `pact-orchestrator` but its `agentId` would not match the team's `leadAgentId`, so the leadAgentId check correctly treats its tasks as teammate work; an `agentType`-based check would silently exclude them and break the 1→0 teardown signal.

## Failure-mode posture

When the team config is unreadable (empty members list), the check short-circuits to \"count\" — under-arm (silent teardown loss while teammate work is in flight) is unrecoverable; over-arm (extra empty scans) is recoverable on the next state change. The asymmetry is documented in an inline comment block inside the step-4 body, structurally anchored by `test_lifecycle_relevant_preserves_fail_conservative_audit_anchor`.

## Test plan

- [x] `pytest pact-plugin/tests/test_wake_lifecycle*.py pact-plugin/tests/test_inbox_wake_lifecycle_helper.py`: 88 passed / 2 skipped
- [x] Broader `pytest pact-plugin/tests/ --ignore=pact-plugin/tests/telegram_notify`: clean (no fix-743 regressions; pre-existing failures in `test_telegram_voice.py` and `test_variety_scorer_fuzz.py` confirmed present on `main` HEAD)
- [x] Counter-test-by-revert: surgically delete the step-4 body in `_lifecycle_relevant` (keep helpers + imports + docstring), clear `__pycache__`, run pytest → empirical 8-fail cardinality (4 dedicated edge cases + 3 legacy lockstep inversions + 1 audit-anchor) matches the cardinality breakdown documented in the new test module's docstring exactly. Restored cleanly via backup.
- [x] Adversarial-edge: stringified non-string owners, malformed/null/missing member fields, empty/null leadAgentId, name-only-and-agentId-only mismatches — all behaviorally pinned.
- [x] Perf: 100×20 stress test runs in <0.1s (5s budget). `_iter_members` invocations bounded at 3× per predicate call; no quadratic regression in `count_active_tasks`.
- [x] Integration: realistic comPACT layout (1 unowned feature task + 2 teammate work tasks + 1 secretary signal) walks count_active_tasks across the full lifecycle, reaching 0 after teammate completion.

## Non-goals

- No changes to workflow command bodies (orchestrate, comPACT, peer-review). The fix lives entirely in the hook predicate.
- No introduction of a `metadata.umbrella=true` marker. Fix B from the issue body was considered and rejected — depends on every workflow command remembering to mark umbrella tasks; brittle.
- No optimization of the 3× `_iter_members` invocation per predicate call (acknowledged in spec §9 risk row; safe to defer).

## Follow-ups filed during this PR

- #751 — scan-pending-tasks warmup-grace skip for the first ~120s after arm (eliminates the immediate-after-arm empty fire).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.2.2",
+      "version": "4.2.3",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.2.2/      # Plugin version
+│               └── 4.2.3/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.2.2",
+  "version": "4.2.3",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.2.2
+> **Version**: 4.2.3
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/hooks/shared/pact_context.py
+++ b/pact-plugin/hooks/shared/pact_context.py
@@ -359,6 +359,53 @@ def _iter_members(
     return [m for m in members if isinstance(m, dict)]
 
 
+def _read_team_lead_agent_id(
+    team_name: str,
+    teams_dir: str | None = None,
+) -> str:
+    """Read the top-level ``leadAgentId`` string from a team config file.
+
+    Returns the ``leadAgentId`` value from
+    ``~/.claude/teams/{team_name}/config.json``, or ``""`` silently on any
+    of the same error paths handled by ``_iter_members``: empty team_name,
+    missing config file, I/O error, malformed JSON, non-object top-level
+    JSON, missing field, or non-string field value.
+
+    Sibling reader to ``_iter_members``. ``_iter_members`` returns only the
+    validated ``members[]`` list (by design — it discards every other
+    top-level field). Rather than widen ``_iter_members``'s return shape
+    (which ripples through its existing call sites), this helper exposes
+    the single additional field that the wake-lifecycle owner-classification
+    predicate needs (``leadAgentId``). Matches the established
+    one-helper-per-field-or-projection pattern in this module.
+
+    Silent-on-error is intentional and parallels ``_iter_members``: the
+    sole caller (``wake_lifecycle._is_lead_owned``) uses the empty string
+    as the "team config not usable" signal and composes that result with
+    the call-site fail-conservative posture.
+
+    Args:
+        team_name: Team name for config path. Empty string returns "".
+        teams_dir: Override teams directory (for testing).
+    """
+    if not team_name:
+        return ""
+    if teams_dir:
+        config_path = Path(teams_dir) / team_name / "config.json"
+    else:
+        config_path = (
+            Path.home() / ".claude" / "teams" / team_name / "config.json"
+        )
+    try:
+        data = json.loads(config_path.read_text(encoding="utf-8"))
+        lead_agent_id = data.get("leadAgentId")
+    except (OSError, json.JSONDecodeError, ValueError, AttributeError, TypeError):
+        return ""
+    if not isinstance(lead_agent_id, str):
+        return ""
+    return lead_agent_id
+
+
 def _lookup_agent_in_team_config(
     agent_id: str,
     team_name: str,
--- a/pact-plugin/hooks/shared/session_journal.py
+++ b/pact-plugin/hooks/shared/session_journal.py
@@ -158,6 +158,16 @@ _REQUIRED_FIELDS_BY_TYPE: dict[str, dict[str, type]] = {
     # activates the _OPTIONAL_FIELDS_BY_TYPE enforcement below (same pattern
     # as session_end and cleanup_summary).
     "session_consolidated": {},
+    # hooks/shared/wake_lifecycle.py::_warn_empty_team_config_once writes
+    # wake_tally_warn ONE-shot per team per process when the step-4
+    # owner-check fall-through fires (config unreadable, members[] empty).
+    # team_name is load-bearing — without it the warn has no actionable
+    # signal. reason is a categorical token for future log-filter
+    # dispatch ("empty_team_config_fail_conservative" today; future
+    # variants may add "malformed_team_config_json" or similar). detail
+    # is documentation-grade prose for human readers (stderr fallback)
+    # and is registered as optional in _OPTIONAL_FIELDS_BY_TYPE below.
+    "wake_tally_warn": {"team_name": str, "reason": str},
 }
 
 
@@ -222,6 +232,15 @@ _OPTIONAL_FIELDS_BY_TYPE: dict[str, dict[str, type]] = {
         "task_count": int,
         "memories_saved": int,
     },
+    # hooks/shared/wake_lifecycle.py::_warn_empty_team_config_once writes
+    # wake_tally_warn with an optional `detail` free-form prose string.
+    # The required-fields registration in _REQUIRED_FIELDS_BY_TYPE above
+    # ("wake_tally_warn": {team_name: str, reason: str}) is what
+    # ACTIVATES this optional check — _validate_event_schema short-
+    # circuits on unknown event types and would otherwise skip the
+    # optional loop. Symmetric with the session_end + cleanup_summary
+    # registration patterns.
+    "wake_tally_warn": {"detail": str},
 }
 
 
--- a/pact-plugin/hooks/shared/wake_lifecycle.py
+++ b/pact-plugin/hooks/shared/wake_lifecycle.py
@@ -108,10 +108,10 @@ class _OwnerClassification:
     - ``agent_type``: the member's ``agentType`` string when present,
       else None. Used by the wake-excluded-agentType carve-out.
     - ``config_readable``: True iff ``_iter_members`` returned a
-      non-empty members list. False distinguishes the
-      fail-CONSERVATIVE-fall-through case (members list empty: config
-      missing, malformed JSON, or members[] absent) from the orphan
-      case (members non-empty, no name match).
+      non-empty members list. False signals the count-on-failure
+      fall-through path at the call site (members list empty: config
+      missing, malformed JSON, or members[] absent) and distinguishes
+      it from the orphan case (members non-empty, no name match).
 
     A frozen dataclass rather than a namedtuple so the field semantics
     are documented inline at the consumer's read site.
@@ -143,7 +143,7 @@ def _classify_owner(owner: Any, team_name: str) -> _OwnerClassification:
       team_name → all fields default (no classification can be made).
     - ``_iter_members`` returns ``[]`` (config missing / unreadable /
       malformed / members[] absent) → ``config_readable=False`` so the
-      call site can fail-CONSERVATIVE (count toward tally). The other
+      call site can apply its count-on-failure fall-through. The other
       fields stay False / None.
     - members non-empty, no name match → ``is_known_team_member=False``,
       ``is_lead=False``, ``agent_type=None``, ``config_readable=True``.
@@ -164,23 +164,24 @@ def _classify_owner(owner: Any, team_name: str) -> _OwnerClassification:
         return _OwnerClassification(False, False, None, False)
 
     # Read the config regardless of owner shape so the call site can
-    # distinguish "config unreadable → fail-CONSERVATIVE" from "config
-    # readable but owner is non-string / orphan → exclude". A bad-owner
-    # task under a readable config is an orphan (excluded), not a
-    # config-read failure.
+    # distinguish "config unreadable → count-on-failure fall-through"
+    # from "config readable but owner is non-string / orphan →
+    # exclude". A bad-owner task under a readable config is an orphan
+    # (excluded), not a config-read failure.
     members = _iter_members(team_name)
     if not members:
-        # Empty members list = fail-CONSERVATIVE signal at the call site
-        # (members[] empty ⇒ unreadable for classification purposes).
-        # See the inline comment block in _lifecycle_relevant for the
-        # under-arm-unrecoverable vs over-arm-recoverable asymmetry.
+        # Empty members list signals the count-on-failure fall-through
+        # at the call site (members[] empty ⇒ unreadable for
+        # classification purposes). The rationale for that posture
+        # lives in _lifecycle_relevant's step-4 elif body where the
+        # logic is exercised.
         return _OwnerClassification(False, False, None, False)
 
     if not isinstance(owner, str) or not owner:
         # Config IS readable, but the owner is missing or non-string.
         # Return is_known=False so the call site excludes the task as
         # an orphan / unowned; config_readable=True signals "the
-        # exclusion is intentional, not fail-CONSERVATIVE."
+        # exclusion is intentional, not the count-on-failure path."
         return _OwnerClassification(False, False, None, True)
 
     lead_agent_id = _read_team_lead_agent_id(team_name)
@@ -218,10 +219,12 @@ def _owner_is_known_team_member(owner: Any, team_name: str) -> bool:
     non-string owner, empty owner string, empty team_name, empty members
     list (config unreadable or members[] missing/empty), or no name
     match. Fail-CLOSED-symmetric internally — the call site
-    (``_lifecycle_relevant``) is responsible for the fail-CONSERVATIVE
-    short-circuit when the members list is empty (see the inline comment
-    block in ``_lifecycle_relevant``'s step-4 body for the asymmetry
-    rationale).
+    (``_lifecycle_relevant``, step 4) handles its own count-on-failure
+    posture when the members list is empty; see the inline rationale
+    comment inside the step-4 ``elif team_name:`` branch for the
+    asymmetry between this helper's fail-CLOSED behavior and the call
+    site's fall-through. Step labels match the ``# Step N:`` anchors
+    in ``_lifecycle_relevant``.
 
     Retained as a public-style helper for test reuse and future callers;
     after the ``_classify_owner`` refactor, ``_lifecycle_relevant`` uses
@@ -239,10 +242,12 @@ def _is_lead_owned(owner: Any, team_name: str) -> bool:
     non-string owner, empty owner string, empty team_name, empty
     leadAgentId (config unreadable or field missing), empty members
     list, or no member matches both name and lead-agentId. Fail-CLOSED-
-    symmetric internally — the call site (``_lifecycle_relevant``) is
-    responsible for the fail-CONSERVATIVE short-circuit when the members
-    list is empty (see the inline comment block in
-    ``_lifecycle_relevant``'s step-4 body for the asymmetry rationale).
+    symmetric internally — the call site (``_lifecycle_relevant``,
+    step 4) handles its own count-on-failure posture when the members
+    list is empty; see the inline rationale comment inside the step-4
+    ``elif team_name:`` branch for the asymmetry between this helper's
+    fail-CLOSED behavior and the call site's fall-through. Step labels
+    match the ``# Step N:`` anchors in ``_lifecycle_relevant``.
 
     Retained as a public-style helper for test reuse; after the
     ``_classify_owner`` refactor, ``_lifecycle_relevant`` uses the
@@ -416,21 +421,10 @@ def _lifecycle_relevant(task: Any, team_name: str = "") -> bool:
     # so a teammate-owned signal task passes step 4 and is excluded at
     # step 6; an unowned/orphan/lead-owned signal task (structurally
     # impossible today, but defended against) is excluded at step 4.
-    #
-    # Fail-CONSERVATIVE: if the team config is unreadable (members list
-    # is empty), `_classify_owner` returned `config_readable=False` and
-    # BOTH step 3 and step 4 fall through to step 5. The sibling
-    # predicates in intentional_wait.py fail-CLOSED on read errors
-    # (return False), but the failure mode here inverts the priority:
-    # under-arm (silent teardown loss while teammate work is in flight)
-    # is unrecoverable; over-arm (extra empty scans) is recoverable on
-    # the next state change. The wake-mechanism's purpose — never
-    # strand a teammate whose SendMessage needs to wake the lead — is
-    # load-bearing here, so we fail toward counting on every config-
-    # read failure. The audit-anchor invariant test pins these phrases
-    # (Fail-CONSERVATIVE, under-arm, unrecoverable) inside this body so
-    # a future agent-reader deleting step 4 also deletes the rationale
-    # rather than leaving an orphan comment.
+    # The failure-mode rationale documenting why the unreadable-config
+    # branch counts (rather than excluding) lives inside the `elif`
+    # body below so a body-only revert deletes both the logic and its
+    # justification together.
     if classification.config_readable:
         if not isinstance(owner, str) or not owner:
             return False
@@ -439,8 +433,23 @@ def _lifecycle_relevant(task: Any, team_name: str = "") -> bool:
         if classification.is_lead:
             return False
     elif team_name:
-        # Non-empty team_name but empty members list — fail-CONSERVATIVE
-        # fall-through (see the asymmetry comment above this block).
+        # Fail-CONSERVATIVE: the team config is unreadable (members
+        # list is empty), so `_classify_owner` returned
+        # `config_readable=False` and BOTH step 3 and step 4 fall
+        # through to step 5. The sibling predicates in
+        # intentional_wait.py fail-CLOSED on read errors (return False),
+        # but the failure mode here inverts the priority: under-arm
+        # (silent teardown loss while teammate work is in flight) is
+        # unrecoverable; over-arm (extra empty scans) is recoverable on
+        # the next state change. The wake-mechanism's purpose — never
+        # strand a teammate whose SendMessage needs to wake the lead —
+        # is load-bearing here, so we fail toward counting on every
+        # config-read failure. The audit-anchor invariant test pins
+        # these phrases (Fail-CONSERVATIVE, under-arm, unrecoverable)
+        # inside this `elif` body so a future agent-reader deleting
+        # step 4 also deletes the rationale rather than leaving an
+        # orphan comment.
+        #
         # One-shot observability warn per team per process so a
         # permanently-unreadable config doesn't silently keep the wake
         # mechanism armed for the entire session.
--- a/pact-plugin/hooks/shared/wake_lifecycle.py
+++ b/pact-plugin/hooks/shared/wake_lifecycle.py
@@ -131,7 +131,11 @@ class _OwnerClassification:
 _EMPTY_CONFIG_WARN_TEAMS: set[str] = set()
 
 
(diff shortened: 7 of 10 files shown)


michael-wojcik (commented):
## Architectural review — fix-743 PR #752

**Reviewer**: pact-architect (spec author — self-review bias acknowledged; pre-diff review angle declared with 5 enumerated bias surfaces + 5 falsifiability-anchor verification commands, per cross-session "Declare review angle BEFORE reading diff" pin).

**Verdict shape**: per-finding severity (BLOCKING / MINOR / FUTURE). Team-lead authorizes merge — this comment does not approve or request-changes.

### Pre-diff falsifiability anchor — verification commands run

| # | Command | Result |
|---|---------|--------|
| 1 | `git log --oneline a88c534d..HEAD` | 6 commits, scopes match PR description (refactor → fix → 2× test → cleanup → version) |
| 2 | `git show aedf77dd -- pact-plugin/hooks/shared/wake_lifecycle.py` | Audit-anchor relocation confirmed: comment moved from BEFORE `if team_name:` to INSIDE the body, structurally tied to the executable branch it counter-signals |
| 3 | `grep -nE "_iter_members\(" pact-plugin/hooks/shared/{wake_lifecycle.py,intentional_wait.py,pact_context.py}` | 3 call sites in `wake_lifecycle.py` (lines 97, 126, 224) — confirms 3× per-task invocation cost; matches PR description acknowledgement and spec §9 risk row |
| 4 | (test diff read via direct file inspection — `gh pr diff --` flag rejected) | 88 passed / 2 skipped; 4 new edge-case tests + 1 behavioral + helpers-pure-never-raises + adversarial + perf + integration |
| 5 | Step-4 truth table from `wake_lifecycle.py:217-232` | 5 branches in correct order: members-empty → fall-through; owner-non-string → False; not-member → False; lead-owned → False; otherwise → step 5. No valid task is trapped. |

### Findings

**(GREEN — design coherence) Implementation matches spec §3.2 step-order exactly.** `pact-plugin/hooks/shared/wake_lifecycle.py:213-232` inserts the owner-check between the existing wake-excluded-agentType check (step 3) and the metadata-shape check (step 5), preserving the signal-task-after-owner ordering rationale (step 4 → step 6 composition). Three new helpers (`_owner_is_team_member`, `_is_lead_owned` at `wake_lifecycle.py:81-133`; `_read_team_lead_agent_id` at `pact_context.py`) have the public shapes the spec specified. Sibling-reader seam chosen (per backend-coder discretion) — does not ripple through `_iter_members`'s three existing call sites. Clean.

**(GREEN — fail-posture asymmetry) Documentation is load-bearing and structurally anchored.** The fail-CONSERVATIVE comment block lives INSIDE the `if team_name:` body (commit `aedf77dd` relocation) rather than in the preamble. This is the correct placement: the comment now structurally tracks the executable branch it documents, and the audit-anchor invariant (`test_lifecycle_relevant_preserves_fail_conservative_audit_anchor` at `test_inbox_wake_lifecycle_helper.py:92`) fires as a true counter-signal under body-only revert. The sibling helpers (`_owner_is_team_member`, `_is_lead_owned`) remain fail-CLOSED-symmetric internally, with both docstrings explicitly noting the call-site asymmetry.

**(GREEN — rePACT forward-compat) Interface contracts hold under the sub-orchestrator scenario.** A `pact-orchestrator`-typed teammate whose `agentId != team_config.leadAgentId` would: pass `_owner_is_team_member` (name matches), fail `_is_lead_owned` (agentId mismatch on the AND-composition at `wake_lifecycle.py:131`), and correctly count toward the wake tally. The AND-composition is explicitly pinned by `test_is_lead_owned_requires_both_name_and_agentid_match`.

**(MINOR — spec drift, post-implementation refinement) Counter-test-by-revert cardinality refined from spec §6.4 (predicted 5 fail) to empirical 8 fail (4 dedicated + 3 legacy lockstep + 1 audit-anchor).** The implementation correctly enumerated three legacy-test lockstep inversions in `test_inbox_wake_lifecycle_helper.py` that my spec did not anticipate (those tests' assertions inverted with the new orphan-exclusion semantics). Logged as MINOR per team-lead's spec-drift guidance — in-repo behavior is correct; the gitignored architecture spec's prediction was incomplete. No remediation requested; documenting for future calibration.

**(MINOR — spec drift, defensive coding the spec did not constrain) `if team_name:` guard at `wake_lifecycle.py:223` is a refinement.** The spec assumed `team_name` always populated. The implementation correctly defensively skips step 4 entirely when `team_name=""` (the default), preserving backward compatibility with any legacy / test caller that omits the argument. This is an improvement on the spec, not a deviation; flagging so future maintainers reading the spec recognize the gap.

**(FUTURE — performance, acknowledged in PR description and spec §9) 3× `_iter_members` per `_lifecycle_relevant` call.** `wake_lifecycle.py:224` calls `_iter_members(team_name)` at the call site, then `_owner_is_team_member` calls it again at line 97, then `_is_lead_owned` calls it a third time at line 126. For a typical workload (50 active tasks × 3 reads = 150 small JSON reads per `count_active_tasks()` invocation) this is negligible; the perf test at `test_wake_lifecycle_teammate_owner_filter.py::test_count_active_tasks_scales_linearly_on_large_team` empirically bounds the cost at <0.1s for 100×20. Path to reduce without compromising clarity: an internal `_classify_owner(owner, team_name) -> Literal["member", "lead", "orphan", "config-unreadable"]` projection inside `wake_lifecycle.py` that does ONE `_iter_members` + `_read_team_lead_agent_id` pass and returns a tagged result; the public helpers become thin wrappers. Defer to a follow-up issue if/when the perf budget tightens; today the cost is sub-threshold.

**(FUTURE — observability, optional) No journal event when step 4 short-circuits to fail-CONSERVATIVE.** If the team config genuinely becomes unreadable mid-session, the wake tally silently over-arms and the only signal is the same empty-scan-fire pattern that motivated this fix. Consider emitting a one-shot journal warning the first time `_iter_members` returns `[]` for a non-empty `team_name` per session. Out of scope for this PR; file as a separate observability follow-up if desired.

### Severity counts

- BLOCKING: 0
- MINOR: 2 (both spec-drift refinements; behavior is correct in-repo)
- FUTURE: 2 (perf optimization; observability hook)
- GREEN observations: 3

### Self-review-bias disclosure

I authored `docs/architecture/fix-743.md`. Bias counter-measures applied (per pre-diff teachback declaration): tests-first read of the diff before source; truth-table construction over step-4 branches before reading the diff prose; explicit grep for spec-undocumented behaviors (the `if team_name:` guard, the AND-composition placement, the audit-anchor relocation target line). No bias-related override of a finding I would otherwise have raised.

Closes: review track for PR #752 from architect angle.

michael-wojcik (commented):
## Test-Coverage Review (test-engineer, self-review-bias-aware)

**Author of commit 8c997471 (TEST-phase) is also this reviewer.** All findings below are anchored to **independent empirical re-verification**, not docstring deference. Counter-measures declared pre-diff: re-execute counter-test-by-revert under MULTIPLE variants, attack own prior HANDOFF claims first.

### Falsifiable verification commands executed (declared pre-diff)

| # | Command | Claim it falsifies | Result |
|---|---|---|---|
| a | `git log --oneline main..HEAD -- pact-plugin/` | Chain shape per PR | 6 commits, ends in 3cd6f31e version bump. **MATCHES** |
| b | Body-only revert via Edit (lines 214-232) + pytest | Docstring 17-fail cardinality | **17 fail / 71 pass / 2 skip — EXACT MATCH** |
| c | `git checkout main -- wake_lifecycle.py pact_context.py` + pytest | Full-section revert behavior | **22 fail / 66 pass / 2 skip — 5 MORE than body-only** |
| d | Stash + run telegram + variety_scorer on main HEAD a88c534d | Pre-existing failure claim | Telegram: 21 fail / 3 pass; variety_scorer: collection error from `NameError: name 'st' is not defined`. **Both pre-exist; my prior HANDOFF claim was correct.** |
| e | `grep -nE '_iter_members\(' wake_lifecycle.py intentional_wait.py` | 3+1=4 per-task call claim | wake_lifecycle has 3 sites (L97, L126, L224); intentional_wait has 3 sites (L206, L246, L290) but only L246 (`_is_wake_excluded_agent_type`) is consumed from `_lifecycle_relevant` step 3. **4 per task confirmed**. |

### Findings

#### YELLOW-1 — Phantom-green classification incomplete in docstring (own commit)

The cardinality docstring in `test_wake_lifecycle_teammate_owner_filter.py` labels 5 tests as "passes either way":

- `test_owner_is_team_member_pure_never_raises`
- `test_is_lead_owned_pure_never_raises`
- `test_is_lead_owned_requires_both_name_and_agentid_match`
- `test_empty_or_missing_leadagentid_does_not_misclassify_lead[]`
- `test_empty_or_missing_leadagentid_does_not_misclassify_lead[None]`

This classification is anchored to **body-only revert only**. Under **full-section revert** (the more aggressive variant — `git checkout main -- wake_lifecycle.py pact_context.py`) all 5 of these tests **FAIL** with AttributeError because the helpers `_owner_is_team_member` / `_is_lead_owned` cease to exist. So they ARE load-bearing counter-signals — under a different revert variant than the docstring documents.

**Recommendation**: Amend the docstring to distinguish body-only-pass vs full-section-pass classifications. Concretely add a third column to the "passes either way" table or split into two subsections. Truly-baseline tests under BOTH revert variants: `test_teammate_owned_task_counted` (case 3) and `test_config_unreadable_fail_conservative` (case 5) only.

Severity: YELLOW (cosmetic for the test contract; baseline classification accurate for its declared scope, but a fresh reader could mistakenly assume those 5 tests are not counter-signal at all).

#### YELLOW-2 — TOCTOU coverage gap on `iter_team_task_jsons` mid-iteration

A fresh LLM reader of `count_active_tasks` would assume the predicate is robust to concurrent task-file modification by a sibling teammate. The actual primitive (`task_utils.iter_team_task_jsons`) does:

```python
for task_file in resolved_team_dir.glob("*.json"):
    ...
    content = task_file.read_text(...)  # race window: file may be deleted/replaced
```

There's no test pinning the behavior under a deletion-mid-iteration race. The fail-open posture means a vanished task file silently degrades to "no count contribution" — operationally recoverable (under-arm is the bug-743 mirror image, but transient) yet under-tested.

Spec §9 acknowledged TOCTOU risk on team-config writes; the same shape applies one layer up at task-file iteration and is not covered. Symlinked-team-config and deeply-nested malformed JSON in member entries are also outside test scope.

**Recommendation**: Follow-up issue (not blocking). Add a TOCTOU-style test that deletes a `*.json` file mid-iteration via a custom generator, asserts the count silently absorbs the missing file rather than raising.

Severity: YELLOW (pre-existing iteration primitive surface, not introduced by this PR; flag rather than block).

#### YELLOW-3 — Two-read race on team config between `_iter_members` and `_read_team_lead_agent_id`

`_is_lead_owned` reads `~/.claude/teams/{team}/config.json` twice via two independent helpers:

1. `_read_team_lead_agent_id(team_name)` — reads `leadAgentId`
2. `_iter_members(team_name)` — reads `members[]`

Between the two reads, the config file could be rewritten by a concurrent process (e.g., team-membership change). The fail-CLOSED-symmetric posture absorbs read failures, but the mismatch case (e.g., `leadAgentId` from old version, `members[]` from new version) is not explicitly tested. The behavior IS deterministic (returns False on no match in the new member list), but the test surface assumes both reads see the same on-disk state.

Issue #437 (`_read_team_config_fields` projection) anticipates folding both reads into a single helper that reads the file once. This PR didn't take that path (sibling reader rationale documented in spec §3.1); the trade-off is explicit but the race-window remains for a future fix.

**Recommendation**: Follow-up issue (low priority). Either document the two-read window inline OR introduce a single-read projection.

Severity: YELLOW (architectural decision documented in spec; surfaces here as a coverage observation, not a bug).

#### GREEN — Pre-existing failure provenance verified

Empirical command (d) confirmed on main HEAD `a88c534d` (no fix-743 code):

```
test_telegram_voice.py: 21 failed / 3 passed
test_variety_scorer_fuzz.py: collection error — NameError: name 'st' is not defined
```

Both pre-exist and are unrelated to fix-743. Prior HANDOFF claim was accurate.

#### GREEN — Perf claim verified by independent grep

3 `_iter_members` call sites in `wake_lifecycle.py` (L97, L126, L224) + 1 in `intentional_wait._is_wake_excluded_agent_type` (L246) consumed by `_lifecycle_relevant` step 3 = **4 calls per task**. Stress test (`test_count_active_tasks_scales_linearly_on_large_team`) confirms 100 tasks × 20 members completes <0.1s on developer hardware. No quadratic regression. Caching opportunity (single team-config-fields projection per `count_active_tasks` invocation) is viable architecturally but out-of-scope here.

#### GREEN — Docstring revert procedure reproducible

Following ONLY the cardinality docstring (no out-of-band context) reproduced the 17-fail cardinality cleanly: `grep -n "if team_name:"` finds L214; delete L214-232 via Edit; clear `__pycache__`; run pytest; restore via `cp` from backup; `git diff --quiet` confirms byte-identity. The byte-precise line range is documented and survives the test as an executable procedure.

### Summary

| Severity | Count |
|---|---|
| RED | 0 |
| YELLOW | 3 |
| GREEN | 3 |

YELLOW-1 is the only finding directly attributable to my own TEST-phase commit. YELLOW-2 and YELLOW-3 are pre-existing surface observations the test plan didn't enumerate; suitable for follow-up issues. The 17-fail cardinality is independently reproducible and matches the docstring exactly. The 22-fail cardinality under full-section revert is a stronger property worth capturing inline.

No counter-vote on merging. Recommend amending YELLOW-1 in a follow-up commit (or accepting as-is and filing the docstring refinement as a follow-up issue).

michael-wojcik (commented):
**Review angle (declared before reading diff)**: Agent-reader comprehensibility audit at the call site. Auditing `_lifecycle_relevant` and the new helpers AS IF I were a future LLM teammate (different agent, no PR/spec context — `docs/architecture/fix-743.md` is gitignored) trying to derive correct mental models from source-level cues alone. The "is the implementation correct" axis is already empirically pinned (8-fail counter-test-by-revert + auditor structural GREEN); marginal review signal lives in the reader-experience axis.

Pre-diff verification commands run: `git log main..HEAD` (6-commit chain confirmed), `gh pr view 752`, `gh api .../comments` (0), `gh api .../reviews` (0), `gh api .../files` (8 files, +888/-34), `gh pr checks` (no checks reported).

---

## Comments (one orthogonal finding + two minor)

### F1 [MINOR — concrete improvement, falsifiability anchor] Call-site `members` prefetch is structurally misleading

In `wake_lifecycle.py::_lifecycle_relevant`, the new step 4 block reads:

```python
if team_name:
    # Fail-CONSERVATIVE: ... (9-line audit-anchor comment)
    members = _iter_members(team_name)
    if members:
        if not isinstance(owner, str) or not owner:
            return False
        if not _owner_is_team_member(owner, team_name):  # internally re-calls _iter_members
            return False
        if _is_lead_owned(owner, team_name):  # internally calls _read_team_lead_agent_id + _iter_members
            return False
    # else: members list empty → fail-CONSERVATIVE; fall through.
```

The local `members` variable is bound and then used ONLY for the `if members:` truthiness check; it is never passed down to the helpers, which independently re-read the config via their own `_iter_members(team_name)` calls. Net behavior is correct (auditor verified), but the shape sends a misleading signal to an agent-reader:

1. **Mental-model mismatch**: a fresh reader sees `members = _iter_members(team_name)` and reasonably expects the bound value to be USED (passed down, iterated, or otherwise consumed). When the variable is silently discarded after one truthiness check, the reader either (a) re-reads the block suspecting a bug, or (b) infers a stale-cache concern and worries about TOCTOU between the local read and the helpers' independent reads. Both inferences cost context.

2. **Real (small) perf cost**: `_iter_members(team_name)` is invoked 3× per `_lifecycle_relevant` call (1 prefetch + 1 inside `_owner_is_team_member` + 1 inside `_is_lead_owned`), each reading + parsing the team config from disk. The PR body acknowledges this in §perf ("3× per predicate call") and the new perf test pins it at 3×, but the 3× count is partly an artifact of the prefetch-then-discard shape, not a hard requirement.

**Concrete improvement**: pick ONE of the two natural shapes —
   - **Option A (no prefetch)**: drop the local `members =` line; check `if not _iter_members(team_name): fall through` directly, OR rely entirely on the helpers' fail-CLOSED-on-empty-members behavior. Reduces calls to 2× (one inside each helper) and removes the misleading-variable smell.
   - **Option B (extend the helpers)**: add an optional `members: list[dict] | None = None` parameter to `_owner_is_team_member` and `_is_lead_owned`. Pass the prefetched `members` down. Reduces calls to 1× per `_lifecycle_relevant` invocation, formalizes the "config snapshot" semantics, and the prefetch becomes meaningful.

Either resolves the agent-reader cost. Option A is the smaller change; Option B is the bigger win on perf. Both are out-of-scope for THIS PR (the fix has landed and is correct), but worth a follow-up issue.

### F2 [NIT] "step 4" cross-references dangle — no visual step labels in source

The helper docstrings (`_owner_is_team_member`, `_is_lead_owned`) and the `_lifecycle_relevant` docstring reference "the inline comment block at step 4 of `_lifecycle_relevant`" 3 times. The source has only ONE "step 4" label (in the comment "Teammate-owner check (step 4): ...") — there is no step-1 / step-2 / step-3 / step-5 / step-7 scaffolding anywhere. A future LLM reader navigating from helper docstring → step 4 succeeds, but a reader navigating bottom-up (start at step 4, find the other steps) is stranded. Either add `# Step N:` labels for the existing carve-outs (~4 lines total) OR rephrase the docstring refs as "see the fail-CONSERVATIVE comment block in `_lifecycle_relevant`" (3 places). Not blocking.

### F3 [NIT] Fail-CONSERVATIVE comment scope conflates two empty-input cases

The 9-line `# Fail-CONSERVATIVE:` audit-anchor block lives INSIDE `if team_name:` and documents the empty-`members`-list case. But there are TWO empty-input failure modes the step-4 check guards against:
- empty `team_name` (default-arg case, also the call site from `count_active_tasks` when team_name="" is passed) → skips step 4 entirely via the outer `if team_name:` ⇒ falls through ⇒ counts.
- non-empty `team_name` + empty `members` (config unreadable) → skips owner-check via the inner `if members:` ⇒ falls through ⇒ counts.

Both paths land at the same fall-through behavior, but the audit-anchor comment is structurally INSIDE the `if team_name:` body, so an agent-reader investigating "what happens when team_name is empty?" reads only the OUTER `if team_name:` guard and never sees the rationale. The empty-team_name case is currently undocumented in the call site, even though it's the path most consumers (test fixtures, single-task `_lifecycle_relevant(task)` calls with default args) actually hit.

Suggested 1-line fix: add a comment above `if team_name:` saying "Empty team_name falls through to step 5 with the owner-check skipped — same fail-CONSERVATIVE posture as empty members below." Documents the parallel path; binds both cases to one audit anchor by reference.

---

## Bias-discipline self-check

The orthogonal-finding requirement is met: F1 was hunted under the agent-reader lens specifically (the prefetched-`members` smell is invisible if you're auditing for "is the predicate correct" — auditor's lens — because the predicate IS correct). I deliberately did NOT re-check semantic correctness (8-fail counter-test-by-revert + auditor GREEN already pinned that axis).

None of these are blocking. Recommend filing F1 as a follow-up issue; F2/F3 as inline cleanup in a subsequent PR.

— backend-coder (author; self-review under bias-counter-measure protocol)

Issue #866, opened 2026-05-29
test_inbox_wake_lifecycle_helper.py: fail_conservative audit-anchor fails on main — 'Fail-CONSERVATIVE' leaked outside step-4 window (wake_lifecycle.py:868)


michael-wojcik:
## Summary

`test_inbox_wake_lifecycle_helper.py::test_lifecycle_relevant_preserves_fail_conservative_audit_anchor` **fails on `main` HEAD** (`9deee6bd`), both **in isolation** and in full-suite runs. Pre-existing; surfaced during the #864 Phase-1 baseline run (it is unrelated to #864 and outside that PR's additive scope).

## What the anchor checks

The test pins that the phrases `"Fail-CONSERVATIVE"`, `"under-arm"`, `"unrecoverable"` appear in `hooks/shared/wake_lifecycle.py` **only inside** the step-4 `elif team_name:` comment-block window — an anti-vacuity guard so a future contributor can't delete the elif body's rationale and leave the assertions passing on a stray docstring occurrence.

## The failure

`"Fail-CONSERVATIVE"` now appears at **`wake_lifecycle.py:868`** — a *different* function's comment (`# Fail-CONSERVATIVE wrap: the "Pure function; never raises" docstring ...`), which is **outside** the validated window. The `phrase not in src_outside_window` assertion therefore fails.

Occurrences found:
- L563 — inside the step-4 window (intended)
- L575 — the self-referential comment listing the phrases
- **L868 — outside the window → trips the anchor**

## Fix options (for the fixer to decide)

1. Reword the L868 comment so it doesn't use the literal `"Fail-CONSERVATIVE"` token (cheapest), or
2. Relax/rescope the anchor — it may be over-broad: it forbids the phrase across the *whole* module rather than just guarding the intended rationale block.

## Process note

`main` is shipping non-green for this test — worth understanding why it merged red (CI gap on this suite, or the L868 comment landed in a PR whose CI didn't run it). Likely introduced by the OPERATIONAL-LULL fail-conservative work in the wake_lifecycle subsystem.

## Relevance

This lives in the cron/wake subsystem that #864 plans to delete in Phase 3 — so it may be mooted there, but it should not stay red on `main` in the meantime.

## Related
- #845 — `test_merge_guard.py` 19 order-dependent failures (the *other* part of the 20-failure baseline; distinct root cause = cross-test state pollution)

Issue #864, opened 2026-05-28
Architectural: remove cron wake machinery — tmux teammateMode delivers natively (phased deletion + in-process safeguard)


michael-wojcik:
## Summary

The original "platform-native wake subsumes the cron → remove it" thesis is **confirmed — with the correct mechanism: tmux `teammateMode`** (separate-process teammates over the bridge / event-driven push). The empirical picture:

- **tmux mode:** ~6 s teammate→lead delivery, **idle-independent, stall-free** (n=8, idle 0→20 min; min 6 s / mean 10.75 s / median 6.5 s; one 34 s tail event, never recurred).
- **in-process mode** (current GrowthBook default): delivery is reactive (`useSyncExternalStore` on `mailbox.revision`) + the React inbox poller; it **stalls when the lead is idle** (~44 s mean on short idle, multi-minute stalls on long idle) because the reactive subscription needs the render loop to tick.

The cron wake machinery is a ~30k-LOC ecosystem (3 slash commands + 3 hooks + the wake-inbox marker pipeline + ~18–32k LOC of tests + protocol sections + CLAUDE.md pins) carrying recurring patch debt (OPERATIONAL-LULL, FIRST-OBSERVABLE-WRITE, LLM-error-mode composites, the #849 scan-arm race, …). That is the `db6efbc0` **DEFENSIVE-CODE-AGAINST-UNSTABLE-SUBSTRATE** pattern — userland defensive code papering over the in-process path's slowness. tmux is the clean exit the pattern predicts. The cron's *entire* marginal value collapses to one sliver: an in-process-default user running unattended who won't switch modes or nudge — served far more cheaply by a startup warning than by maintaining the ecosystem.

## Plan (phased)

Full implementation plan: `docs/plans/remove-cron-machinery-plan-2026-05-28-v4.md`.

0. **Broaden the tmux evidence** beyond n=8 / one machine (a realistic full `/PACT:orchestrate` + `/PACT:peer-review` cycle, ≥1 more environment) — **go/no-go gate** before any deletion.
1. **Land a lightweight in-process safeguard** *before* deletion: a `session_init` startup notice (when effective `teammateMode == in-process`) recommending `--teammate-mode tmux` for unattended runs + a documented one-line DIY heartbeat.
2. **Disable cron auto-arm** behind a sentinel constant (both emit sites: `wake_lifecycle_emitter.py` PostToolUse + `session_init.py` resume) — validation window; cron code stays on disk.
[...]

Issue #845, opened 2026-05-27
test_merge_guard.py: 19 order-dependent failures in full-suite runs (cross-test state pollution)


michael-wojcik:
## Summary

`pact-plugin/tests/test_merge_guard.py` has 19 tests that fail when run as part of the full pact-plugin test suite but pass when run in isolation. Pattern is cross-test state pollution — the failing tests depend on shared state that earlier tests modify and don't restore.

## Evidence

| Run mode | Result |
|---|---|
| Standalone (`pytest tests/test_merge_guard.py`) | **975 passed / 0 failed** |
| Full suite (`pytest tests/`) | **956 passed / 19 failed** (in `test_merge_guard.py`) |

Empirical diagnosis surfaced during PR #844 verification (cycles 1–3) and again during PR #844 blind-wave review at parent-baseline commit `18edb7bd`. Same 19 failures + same delta across both parent and merged state → unrelated to #841 work; pre-existing.

## Suspected failing tests (from full-suite run at v4.3.1)

```
TestGhPrCloseAuthorization::test_gh_pr_close_delete_branch_consumes_token
TestGhPrCloseAuthorization::test_merge_token_cannot_authorize_close
TestGhPrCloseAuthorization::test_close_token_cannot_authorize_merge
TestGhPrClosePreHookE2E::test_pre_hook_allows_gh_pr_close_delete_branch_with_token
TestGhGlobalFlagBypassEdgeCases::test_flagged_merge_authorized_with_matching_token
TestLegacyTokenBoundary::test_legacy_token_blocks_second_command_end_to_end
TestLegacyTokenBoundary::test_legacy_token_near_ttl_boundary
```
(plus 12 more — full list visible at any full-suite run head)

## Suspected root cause

The merge_guard module persists token state to a file (or env-var, or shared module global). When a prior test mutates this state without proper teardown, subsequent tests see polluted state and fail their token-presence/absence assertions.

## Suggested fix paths

1. **Add per-test isolation** via pytest `tmp_path` fixture for the token-storage file, OR a `monkeypatch` for the module-level state.
2. **Identify the polluting test** by bisecting: run `test_merge_guard.py` after each other test directory to find which suite leaves the polluted state behind.
3. **Use pytest-randomly + pytest-order** to make the failure pattern reproducible and locate the responsible cross-suite dependency.

## Why now

[...]

PR #867, opened 2026-05-29
feat(#864): in-process teammateMode safeguard (Phase 1)

michael-wojcik:
## Summary

**Phase 1 of #864** — the phased removal of the cron wake machinery. This phase is **purely additive**: it lands the lightweight in-process safeguard **before** any deletion. No cron/wake machinery is touched (that is Phases 2–3).

The empirical conclusion from the #864 investigation: in-process `teammateMode` (the current default) stalls wake delivery when the lead is idle, while tmux mode delivers natively. Rather than keep patching the cron ecosystem, the plan steers users to tmux and covers the residual in-process sliver with a cheap startup notice.

## What landed

- **`pact-plugin/hooks/shared/teammate_mode.py`** (new) — fail-open resolver mirroring the Claude Code runtime getter's file precedence: project `.claude/settings.local.json` → project `.claude/settings.json` → `~/.claude/settings.json` → `~/.claude.json` (legacy) → `'auto'`. `should_emit_inprocess_notice()` returns `True` unless the resolved mode is exactly `'tmux'`. Both public functions are **total** (never raise).
- **`pact-plugin/hooks/session_init.py`** — emits a one-time startup notice (via `system_messages`) when the effective mode is non-tmux, gated on SessionStart `source in {startup, resume}` (suppressed on `compact`/`clear`). Lazy import + belt-and-suspenders guard so the bootstrap path never raises.
- **`pact-plugin/reference/unattended-runs.md`** (new) — recommends `--teammate-mode tmux` for unattended runs, the permanent `~/.claude/settings.json` option, and a one-line DIY heartbeat.
- **Tests** — `test_teammate_mode.py` (+29) and `test_session_init.py` (+8): the §9.1 precedence/should_emit matrix including the **false-suppress edge** (project-local `in-process` over user `tmux` must resolve `in-process` and EMIT), fail-open under malformed/unreadable/wrong-type input, the §9.2 emit-gate integration matrix, and a belt-and-suspenders test.

## Design invariant

**NEVER-FALSE-SUPPRESS.** A false suppress silently reinstates the #864 in-process idle-stall; a false emit is one harmless startup line. Every fail-open path resolves toward EMIT, and the FULL (not two-file) precedence is what closes the false-suppress edge. The tests pin this with counter-test-by-revert: reverting the notice fails the present-rows, reversing precedence fails the false-suppress edge, and inverting the emit polarity fails broadly.

## Key correction surfaced during PREPARE

The plan's open question framed `teammateMode` as a GrowthBook flag. It is **not** — it is a first-class Claude Code **setting** (`~/.claude/settings.json` authoritative, `~/.claude.json` a legacy fallback), and the CLI `--teammate-mode` override is **in-memory-only / hook-invisible**. The fail-safe absorbs the CLI blind spot (unreadable → unknown → emit).

## ⚠️ Pre-existing test baseline (NOT introduced by this PR)

`main` currently ships with **20 full-suite test failures**, unrelated to this change:
- 19 in `test_merge_guard.py` — order-dependent cross-test state pollution (passes in isolation). Tracked: **#845**.
- 1 in `test_inbox_wake_lifecycle_helper.py` — `fail_conservative` audit-anchor (`'Fail-CONSERVATIVE'` leaked outside the step-4 window at `wake_lifecycle.py:868`). Tracked: **#866** (filed during this work).

**This PR adds ZERO new failures** — proven via sorted node-id delta (baseline 20 == post-impl 20, identical sets). Full suite: 8402 passed / 20 failed / 14 skipped, +37 new tests all green. CI reds on these 20 are pre-existing; do not attribute them to this PR.

## Version

PATCH `4.3.4` → `4.3.5` (additive-only phase; no user-facing command-surface changes).

## Phasing

This is Phase 1 of 3. Phase 2 (disable cron auto-arm behind a sentinel, validation window) and Phase 3 (delete the ecosystem) are gated on this baking + cross-environment tmux validation. Plan: `docs/plans/remove-cron-machinery-plan-2026-05-28-v4.md`.

Closes (partial): #864 (Phase 1 only).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.3.4",
+      "version": "4.3.5",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.3.4/      # Plugin version
+│               └── 4.3.5/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.3.4",
+  "version": "4.3.5",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.3.4
+> **Version**: 4.3.5
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/hooks/session_init.py
+++ b/pact-plugin/hooks/session_init.py
@@ -6,6 +6,7 @@ Used by: Claude Code settings.json SessionStart hook
 
 Performs PACT environment initialization:
 0. Checks if ~/.claude/teams is in additionalDirectories (emits setup tip if not configured)
+0b. Emits a one-time in-process teammateMode notice recommending tmux for unattended runs (startup/resume only)
 1. Creates plugin symlinks for @reference resolution
 3. Ensures project CLAUDE.md exists with memory sections
 3b. One-time migration: wraps existing project CLAUDE.md in PACT_MANAGED boundary (#404)
@@ -100,6 +101,19 @@ from shared.session_resume import (
     restore_last_session,
     check_resumption_context,
     check_paused_state,
+)
+
+
+# #864 Phase 1: one-time startup notice recommending tmux for unattended runs
+# when the effective teammateMode is not positively "tmux". Emitted via
+# system_messages (user-facing) by main() step 0b. Lives HERE (presentation
+# layer) rather than in shared/teammate_mode.py (resolution layer) per SRP.
+# Pure literal (no interpolation) so tests can pin the exact substring.
+_INPROCESS_MODE_NOTICE = (
+    "PACT: unattended runs may stall in in-process teammate mode "
+    "(the lead can sit idle awaiting a wake that needs a manual nudge). "
+    "For hands-off runs, relaunch with `--teammate-mode tmux` for reliable "
+    "native delivery, or keep a heartbeat — see reference/unattended-runs.md."
 )
 
 
@@ -572,6 +586,7 @@ def main():
 
     Performs PACT environment initialization:
     0. Checks if ~/.claude/teams is in additionalDirectories (emits setup tip if not configured)
+    0b. Emits a one-time in-process teammateMode notice recommending tmux for unattended runs (startup/resume only)
     1. Creates plugin symlinks for @reference resolution
     3. Ensures project CLAUDE.md exists with memory sections
     3b. One-time migration: wraps existing project CLAUDE.md in PACT_MANAGED boundary (#404)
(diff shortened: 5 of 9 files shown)


michael-wojcik:
## Post-review remediation (all reviewers GREEN, zero blocking)

3 fresh independent reviewers approved. 3 MINOR findings fixed in this PR; 1 FUTURE filed as an issue:

- **M1** (`be199141`) — corrected a shipped docstring that falsely claimed omitting enterprise managed-settings is "over-emit only / safe direction"; it's actually a narrow false-suppress, now named as the 2nd of exactly two accepted never-false-suppress breaches (1st = in-memory CLI override). Code fix tracked by **#868**.
- **M2 + M3** (`47dc9021`) — anti-fragility regression tests pinning two phantom-green refactor traps that both reviewers independently flagged: a `.strip().lower()` normalization (would flip case/whitespace variants to SUPPRESS) and a narrow-except refactor (would let a non-OSError/JSONDecodeError escape the per-file read). Counter-tested: each mutation fails the new tests.
- **F2 + F3** folded into `be199141` (allowlist-maintenance comment + Phase-2 reuse-contract docstring note).

Note: the M3 discriminator uses invalid-UTF-8 bytes, not a NUL path — empirically, `Path.exists()` returns False for a NUL path on CPython 3.14.5, so a NUL test would be phantom-green.

Full suite: 8412 passed / 20 pre-existing failures (#845 + #866, unrelated) / 14 skipped — **zero new failures**.

michael-wojcik:
## F1 folded in (managed-settings false-suppress closed in code)

Per the post-review feasibility spike (clean GO — confirmed OS-specific managed-settings paths from the live runtime getter), F1 was folded into this PR rather than deferred:

- **`feat(#864): read enterprise managed-settings at top precedence`** (`49ef18b6`) — `teammate_mode.py` now reads the OS-specific enterprise managed-settings.json as the **highest-precedence** source. A managed `teammateMode` of in-process/auto over a lower-layer tmux now resolves correctly (managed → non-tmux → emit) instead of false-suppressing. Fail-open preserved (absolute literal read via the total `_read_teammate_mode`; absent on dev machines → skip → zero behavior change off managed fleets). The only remaining accepted never-false-suppress breach is the in-memory `--teammate-mode` CLI override (hook-invisible).
- Tests: per-OS path coverage + a managed-precedence suite including the **breach-closure regression** (managed in-process over user tmux → emit) and its symmetric (managed tmux over user in-process → suppress, also removing managed-tmux over-notification). Counter-test-proven coupled to the precedence prepend.
- Docstrings (source + test + public `resolve_*`) updated: managed-settings now read + tested; **Closes #868**.

Final: 6 commits, full suite zero new failures beyond the pre-existing baseline (#845 + #866). devops-reviewer GREEN (10-probe managed-path variant, 0 raises).

michael-wojcik:
## Round 2 — blind review (complete)

A second review wave with 3 **fresh blind reviewers** (no access to Round-1 findings), on the full current PR:

- **devops-blind** — ✅ APPROVE. Fail-open verified empirically: 0 raises across 33 hostile inputs; four independent fail-open layers; precedence faithful; strictly additive.
- **test-blind** — 🟡 YELLOW → resolved. Caught a real determinism gap Round 1 missed: the default test fixture neutralized `Path.home`/`CLAUDE_PROJECT_DIR` but not `_managed_settings_path`, so ~24 tests read the real OS managed path — non-deterministic on a managed fleet (F1's own target env). Fixed in `eb1e0cc4` (fixture isolation + a binding companion + the managed-auto breach-closure case). All 4 invariant discriminators independently confirmed *genuine* (not phantom-green).
- **architect-blind** — 🟢 GREEN (fidelity-audit complement; disclosed it cannot be truly blind since its agent-type memory authored this design). Confirmed the M1 reframe, precedence (re-derived from source), and docstring byte-parity are all clean. Noted a stale **gitignored** design doc (accepted as superseded — not shipped) and the version-tier observation.

Final: 7 commits, full suite **8422 passed / 20 pre-existing failures (#845 + #866) / 14 skipped — zero new failures**.

Net: the code is independently confirmed sound, and the one real finding (test determinism) is fixed.

michael-wojcik:
## L3 + L4 folded in — all review findings now addressed

The two previously-deferred LOW coverage notes are now pinned, plus the session_init-side twin of the test-determinism fix:

- **L3** (`78f3c5e5`) — unrecognized SessionStart source → suppress (test_unrecognized_source_suppresses_notice).
- **L4** (`78f3c5e5`) — missing source key → defaults to 'startup' → emit (test_missing_source_defaults_to_startup_and_emits).
- Hardened the shared notice-integration helper to isolate `_managed_settings_path` (the session_init-side twin of the Round-2 fixture fix) — determinism hardening now complete across both test files; the 8 existing helper-using tests verified behavior-neutral.

Counter-test-by-revert confirmed both new tests are genuine orthogonal discriminators. Final: 8 commits, full suite **8424 passed / 20 pre-existing failures (#845 + #866) / 14 skipped — zero new failures**. Every finding from both review rounds (Round 1 + the blind Round 2) is now resolved.

PR #952, opened 2026-06-13
fix(hooks): add content-assertable pactGateHealth crash-path marker to task_lifecycle_gate (#951)

### LINKED ISSUE #951: task_lifecycle_gate self-masks import crashes (exit 0) — add content-assertable health marker
## Summary

`task_lifecycle_gate.py` still self-masks import crashes: its top-level `except BaseException` handler emits a failure advisory but exits 0, so nothing distinguishes a healthy run from a crashed one at the process level. This is the last of the three self-maskers identified during the #941 PREPARE audit (see #943's comments) — `bootstrap_marker_writer` and `bootstrap_prompt_gate` gained context self-heal + content-asserted tests in #950, but `task_lifecycle_gate`'s masking is **structural**: it is a PostToolUse hook, and PostToolUse hooks cannot deny, so fail-closed is not available as a posture.

Surfaced as a Future finding by the PR #950 review panel (review-test F5); user opted to track it as its own issue + PR rather than expand #950's reviewed surface.

## Design constraints (from the #950 cycle)

1. **PostToolUse cannot deny** — the fix cannot be a permission decision; it must be a *visibility* mechanism.
2. **Health must be asserted on output content, never return codes** — exit codes are uninformative by construction here (the masking pattern is `except BaseException → advisory + exit 0`).
3. The hook's consumers (journal emitters, lifecycle bookkeeping) silently lose events when it crashes — the failure mode is **silent journal/event loss**, not blocked tools.

## Proposed direction (sketch)

Emit a structured, machine-recognizable failure marker in the hook's output on the crash path, so downstream consumers (and tests) can discriminate:

```python
except BaseException as err:  # existing catch
    payload = {
        "hookSpecificOutput": {
            "hookEventName": "PostToolUse",
        },
        "systemMessage": (
            f"PACT task_lifecycle_gate DEGRADED ({type(err).__name__}): "
            "lifecycle events for this tool call were not recorded."
        ),
        # structured marker for tests + telemetry; absent on healthy runs
        "pactGateHealth": {"status": "crashed", "stage": "import-or-runtime",
                            "error": f"{type(err).__name__}"},
    }
    print(json.dumps(payload))
    sys.exit(0)  # exit code unchanged — PostToolUse cannot deny anyway
```

Tests then assert health on the **presence/absence of `pactGateHealth`** (content), with a broken-import subprocess sim reusing the #950 scaffold (`_run_degraded_subprocess` pattern in `test_bootstrap_gate.py`).

Open design questions for ARCHITECT:

- Should the marker ride `systemMessage` only (user-visible) or also a dedicated journal event (durable)? The #943 incident argues for durable.
- Truncation/bounding of `{error}` text per the #950 M1 convention (raw exception text into model-visible output is bounded there).
- Whether the same structured-health pattern should be retrofitted onto the other exit-0 advisory paths for consistency.

## Acceptance criteria

- With imports artificially broken, `task_lifecycle_gate` emits a machine-recognizable health marker (content-assertable), and a subprocess test pins it.
- Healthy-path output unchanged (no marker present).
- Health assertions in tests reference output content, never return codes.

michael-wojcik:
## Summary

Closes #951 — `task_lifecycle_gate.py` self-masked import crashes (`except BaseException` → advisory + exit 0), making a crashed run indistinguishable from a healthy one and causing **silent journal/event loss**. This PR adds a content-assertable health surface on the crash path while keeping the healthy path byte-identical.

## What landed

**C1 — floor marker (`c533c140`)**
- Top-level `pactGateHealth` key `{v:1, hook, status:"failed", stage, error}` emitted from `_emit_load_failure_advisory` (single emission point covers both crash stages: module-import and runtime)
- `systemMessage` mirror of the bounded advisory (degraded-visibility convention parity)
- `_bounded_error_text` (cap 200) copied verbatim from `bootstrap_gate.py`; also fixes the previously **unbounded** error interpolation in the existing advisory
- Floor is stdlib-pure, stdin-free, printed first; stderr keeps full raw error text

**C2 — best-effort journal bonus (`3e888d5f`)**
- `_emit_gate_health_event`: guarded late emit of a `gate_health` journal event on the crash path
- Guard wraps the late **imports themselves** — empirically grounded: a live probe showed the fat `shared/__init__` package init fails the late import at package level on `pact_context`-class breakage, so an interior-only guard would let the crash handler crash
- Event type deliberately unregistered (never-load-bearing channel; zero consumers today); two distinct stderr disposition lines discriminate `skipped` vs `unavailable`

**C3 — init move-in (`de233103`)**
- `pact_context.init` moved inside the runtime try: the file's last unmasked crash path now routes through the advisory (marker + journal event)

**TEST (`a12138f1`)** — 12 new legs (suite 8861 → 8873, 0 failed, 0 errors):
- New `tests/test_task_lifecycle_gate_degraded.py`: 3-vector broken-import subprocess matrix asserting full marker content on raw stdout; healthy-path byte-identity pins (the suite's first key-ABSENCE asserts); py3.9-floor leg run live
- In-process legs: runtime-stage marker, the journal channel's one pin (exact 8-key `gate_health` event, fully sandboxed), C3 routing pin, bounded-error truncation leg
- Non-vacuity proven per commit via file-state checkpoints in a detached worktree: pre-C1 → 9 legs RED; C1-only → 3 RED; C1+C2 → 1 RED
- All health assertions on output **content**, never return codes (issue acceptance criterion)

**Version (`624a6928`)** — PATCH bump 4.4.19 → 4.4.20 (canonical 4 files)

## Why

PostToolUse hooks cannot deny, so fail-closed isn't available — visibility is the only posture. Exit codes are uninformative by construction on this path; the platform's stdout schema strips unknown keys (safe but platform-invisible), so raw-stdout subprocess tests are the assertable channel and stdout is the only both-modes-uniform surface. The journal event is best-effort bonus by design: the breakage vectors that crash the hook are the same ones that can break the journal import path.

## Verification

- Full suite: `8873 passed / 13 skipped / 0 failed / 0 errors` (errors token explicitly checked via raw pytest summary)
- Auditor signal: GREEN (all structural ACs verified against `git diff`; byte-identity of the verbatim helper copy proven by sed-range diff; closure-neutrality of lazy imports proven by import set-diff)
- Empirical grounding: 13-leg isolated live probe (scratch env, zero live-tree mutation) corrected one static feasibility claim before design freeze

## Follow-up (tracked separately, post-merge)

- S6/S8 silent-event-loss visibility gap (per blueprint §7.4 sketch) — to be filed as its own issue, bundling `gate_health` registration + `session_resume` surfacing as candidates

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.4.19",
+      "version": "4.4.20",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -605,7 +605,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.4.19/      # Plugin version
+│               └── 4.4.20/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.4.19",
+  "version": "4.4.20",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.4.19
+> **Version**: 4.4.20
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/hooks/bootstrap_gate.py
+++ b/pact-plugin/hooks/bootstrap_gate.py
@@ -182,13 +182,20 @@ def _bounded_error_text(error: BaseException) -> str:
     spaces, and the result is truncated to _ERROR_TEXT_MAX chars with an
     explicit marker. Full text still goes to stderr at the call site.
 
-    Total over hostile exceptions: rendering the message runs the
-    exception's own __str__ (arbitrary code that can itself raise) — fall
-    back to the type name, which renders for any exception class."""
+    Total over hostile exceptions: BOTH renderings that run exception-owned
+    code are guarded. The type name is captured first — a metaclass can make
+    __name__ a property that raises — falling back to a literal; the message
+    runs the exception's own __str__ (arbitrary code that can itself raise),
+    falling back to a marker. Neither fallback touches the exception again,
+    so the function returns a string for any exception object."""
     try:
-        text = f"{type(error).__name__}: {error}"
+        type_name = type(error).__name__
+    except BaseException:  # noqa: BLE001 — hostile metaclass __name__ must not escape
+        type_name = "exception"
+    try:
+        text = f"{type_name}: {error}"
     except BaseException:  # noqa: BLE001 — hostile __str__ must not escape the renderer
-        text = f"{type(error).__name__}: <exception str() raised>"
+        text = f"{type_name}: <exception str() raised>"
     text = "".join(ch if ch.isprintable() else " " for ch in text)
     if len(text) > _ERROR_TEXT_MAX:
         text = text[:_ERROR_TEXT_MAX] + "...[truncated]"
--- a/pact-plugin/hooks/bootstrap_gate.py
+++ b/pact-plugin/hooks/bootstrap_gate.py
@@ -87,19 +87,34 @@ def _emit_load_failure_deny(stage: str, error: BaseException) -> NoReturn:
     import below fails. Audit anchor: hookEventName must be present in any
     deny output.
     """
+    # Guarded rendering BEFORE the deny print: an unguarded hostile/raising
+    # __str__ here would suppress the deny output and exit nonzero-non-2 —
+    # for PreToolUse that is a non-blocking error and the tool call
+    # PROCEEDS (fail-open). The deny must print for any exception; the
+    # fallback is a raise-proof constant.
+    try:
+        error_render = f"{type(error).__name__}: {error}"
+    except BaseException:  # noqa: BLE001 — hostile __str__ must not suppress the deny
+        error_render = "<error text unavailable>"
     print(json.dumps({
         "hookSpecificOutput": {
             "hookEventName": "PreToolUse",
             "permissionDecision": "deny",
             "permissionDecisionReason": (
                 f"PACT bootstrap_gate {stage} failure — blocking for safety. "
-                f"{type(error).__name__}: {error}. Check hook installation "
+                f"{error_render}. Check hook installation "
                 "and shared module availability."
             ),
         }
     }))
+    # Guarded full-text rendering: a raise here would replace the deliberate
+    # exit 2 (blocking) with a traceback exit 1 (non-blocking → fail-open).
+    try:
+        error_full = f"{error}"
+    except BaseException:  # noqa: BLE001 — hostile __str__; keep the exit-2 path
+        error_full = "<exception str() raised>"
     print(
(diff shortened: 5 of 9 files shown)


michael-wojcik:
## Review outcome — 4-panel blind review + 2 remediation cycles

**Panel verdict:** 4× APPROVE, 0 Blocking (architect / test / backend / security).

All findings were Minor / defense-in-depth and have been resolved in-PR:

| Finding | Reviewer | Resolution |
|---------|----------|------------|
| Guard-breadth asymmetry (`_emit_gate_health_event` `except Exception` narrower than the `BaseException` raise surface it wraps) | architect F1 | `c66f4355` — widened to `except BaseException` |
| Hostile `__str__` could suppress the floor marker | security/backend M1 | `c66f4355` — call-site constant fallback + raise-proof error text |
| Vacuous sanitize assert (control chars past the cap) | test M1 | `c021b32e` — payload reshaped inside cap, counter-tested |
| Import-stage journal-success path unpinned | test M2 | `c021b32e` — `TestImportStageJournalSuccess` |
| 3 `bootstrap_gate` residual error-text renders (incl. a deny-path fail-open shape) | folded | `7830b9ec` — guarded renders, exit-2/exit-0 preserved |
| **`_bounded_error_text` non-total vs a raising `__name__`** → bootstrap degraded path exited 1 + suppressed warning (caught by the verify-only adversarial pass) | security (verify-only) | `7155516d` — capture type name safely once; `6636a59a` — regression pins |

**Verify-only re-review:** all findings confirmed RESOLVED at the tip; `_bounded_error_text` now total against hostile `__str__` *and* `__name__`, byte-identical across both hooks.

**Suite:** `8888 passed / 13 skipped / 0 failed / 0 errors`.

**Tracked for a follow-up issue (out of scope, non-reachable):** S6/S8 silent-event-loss visibility gap (`gate_health` registration + `session_resume` surfacing); pre-existing unguarded-stderr `BrokenPipeError`; `main()` unbounded primary stdin read.

michael-wojcik:
## Round-2 fresh blind panel + totality close-out

A second, independent blind review panel (fresh teammates, no round-1 context) re-reviewed the merge candidate. Verdict: **4/4 APPROVE-class, 0 Blocking** — and it earned its keep by surfacing a deeper residual the first round missed.

### The `_bounded_error_text` totality chain

The first round hardened the crash-path renderer against a raising `__str__`. The fresh panel found the renderer's *fallback branch re-interpolates the type name*, so each fix that patched a specific `__name__` value left the next value able to poison the fallback. Closed progressively, then terminated structurally:

| Cycle | Residual | Fix |
|-------|----------|-----|
| 2 | `__name__` **raises** | guard the access |
| 3 | `__name__` **returns a non-str** | `isinstance` coerce |
| 4 | `__name__` **returns a `str` subclass** with a raising `__format__`/`__str__` (passes `isinstance`) | **exact-type guard `type(x) is str`** — the structural terminus |

**Why cycle-4 terminates it:** `type(x) is str` admits *only* an un-spoofable exact `str`, whose `__format__`/`__str__` are str's unpatchable C builtins — so the fallback cannot raise on the type name for **any** original `__name__` value (including a `__class__`-spoof that defeats `isinstance`). The security reviewer adversarially confirmed no reachable defeat. Generalizable lesson: for reducing a hostile value to a safe primitive, identity (`type(x) is T`) beats behavioral (`isinstance(x, T)`).

All residuals were **non-reachable** (a hostile metaclass needs code execution, never stdin) — the feature was defect-free for all realistic inputs throughout; these were defense-in-depth + totality-claim accuracy.

### Also closed
- `BaseException`-breadth regression pin (narrowing the gauntlet guard would re-mask #951 — now caught by a test)
- Both bootstrap_gate degraded/deny paths verified total under all hostile classes

### Verification
- Suite: **`8909 passed / 13 skipped / 0 failed / 0 errors`**
- Both hooks' `_bounded_error_text` byte-identical (sha256 verified each cycle)
- Every fix counter-tested (source-revert fences with precise bite/green partitions)

### Follow-up issue (pre-existing, non-reachable, tracked separately)
deny-path unbounded error text; `main()` unbounded primary stdin; stderr `BrokenPipeError`; multi-GB exact-str `MemoryError` in the truncation join.

michael-wojcik:
Deferred (pre-existing, non-reachable) follow-up items are tracked separately:
- #953 — pre-existing crash-path / error-render hardening in the gate hooks (deny-path length bound, primary stdin bound, stderr BrokenPipeError, truncation MemoryError)
- #954 — S6/S8 gate_health event observability (registration + session_resume surfacing), the originally-deferred #951 scope

This PR folds all in-scope findings; #953/#954 capture everything explicitly deferred.

PR #959, opened 2026-06-14
fix(hooks): journal-durability hardening — write-after-completion race (#956) + GC-immune variety/ack mirror (#955)

### LINKED ISSUE #955: Mirror per-dispatch variety + teachback-ack into the session journal (survive task-store GC for wrap-up Q5/Q6)
## Problem

The wrap-up Orchestration Retrospective's **Q5 (variety divergence)** and **Q6 (variety-acknowledgment signals)** read their inputs from task metadata:
- Q5 — `metadata.variety.total` across Task-B work tasks
- Q6 — `metadata.teachback_submit.variety_acknowledgment.rationale_articulates_this_dispatch` across Task-A teachback tasks

Both sub-signals are **GC-vulnerable**: when the task store GCs the older tasks before consolidation runs (observed mid-session in a long arc — the store wiped to the two newest tasks), `compute_variety_divergence` returns a `no_dispatches_stamped` / coverage 0.0 result that is a **false-empty** — the tasks *were* stamped, the data is just gone. The session journal mirrors only the **feature-level** variety, not the per-dispatch stamps or the teachback acknowledgments, so there's no durable fallback the way `agent_handoff` events survive GC.

Net effect: the two highest-value *per-dispatch* calibration sub-signals are silently lost on exactly the long, high-dispatch sessions where they'd be most informative.

## Observed instance

A 26-dispatch arc (#951/#952) reached consolidation with the task store GC'd to the two newest tasks. Q5/Q6 came back coverage-0.0 false-empty; the real divergence (feature 8 vs dispatch mean ≈6.58, delta 1.42, within threshold) was only recoverable because the orchestrator happened to still hold the per-dispatch values in its own context and hand-supplied them — a manual salvage that won't always be possible.

## Proposed fix (mitigation b — durable journal mirror)

Mirror the two fields into the session journal **at write time**, so they survive GC like `agent_handoff` does:

- **At dispatch** (Task-B variety stamp): emit a small `dispatch_variety` journal event `{task_id, variety: {novelty, scope, uncertainty, risk, total}}`.
- **At teachback acceptance** (Task-A): emit a `teachback_ack` event `{task_id, rationale_articulates_this_dispatch, concern}`.

Then the wrap-up Q5/Q6 aggregation reads the journal (GC-immune) instead of (or as a fallback to) the task store. `compute_variety_divergence` already takes a plain list — only the *source* changes.

## Interim mitigation (a — secretary-side, no plugin change)

The secretary snapshots per-dispatch variety + teachback-ack during the **incremental** harvest passes (while task JSONs still exist), so consolidation has the data even after GC. Lighter but relies on incremental passes running before GC.

## Notes
- Surfaced during the #951/#952 wrap-up by the secretary (`calibration-subsignal-gc-loss` pattern).
- Low risk: additive journal events; no change to existing schemas or the `compute_variety_divergence` helper signature.

### LINKED ISSUE #956: agent_handoff journal event lost when a task is completed before its metadata.handoff is written
## Problem

The `agent_handoff` journal emitter mirrors a task's `metadata.handoff` into the session journal at **task-completion time**. If a HANDOFF-bearing teammate task is marked `completed` **before** the teammate has written `metadata.handoff`, there is nothing to emit; when the HANDOFF lands moments later, the task is already `completed`, so no re-fire occurs — and the HANDOFF is **never journal-mirrored**.

The journal is the GC-immune record (it survives task-store GC, unlike the task JSON). So a HANDOFF lost from the journal this way is lost from the durable record entirely — recoverable only out-of-band (e.g. via `git`, if the work was committed).

## Observed instance

In a long arc (#951/#952), the session journal carried `agent_handoff` events for 20 tasks but **not** for `#54` (a cycle-4 hook fix). Root cause: the lead marked `#54` completed immediately after committing its staged work, a beat *before* the teammate wrote its `metadata.handoff`. The HANDOFF content was recovered via `git`, so nothing was ultimately lost — but only because the work happened to be committed and the orchestrator still held context.

## Two facets

1. **Procedural (lead-side):** the orchestrator protocol already says *do not mark a task completed if `metadata.handoff` is missing/empty*. Disciplined lead behavior avoids this. (In the observed instance this was a lead slip.)
2. **Plugin-coverage (the real gap):** even with correct discipline, completion and the teammate's HANDOFF-write can interleave in a race. The emitter has **no path** to mirror a HANDOFF that is written *after* completion. This is distinct from the known mid-turn-`TaskCompleted`-skip class (the b2 lead-side emit, v4.4.4) — there `b2` fires at completion but finds **no handoff to emit**.

## Proposed mitigations

**Preferred — prevent the race (gate-side):**
A PreToolUse check on `TaskUpdate(status="completed")` for a teammate-owned, HANDOFF-expecting task: if `metadata.handoff` is missing/empty, **warn or deny** the completion (mechanically enforcing the existing protocol rule). This eliminates the race at the source rather than recovering from it.

**Complementary — write-time backstop (emitter-side):**
On the `TaskUpdate` that *sets* `metadata.handoff`, if the task is already `completed` and no `agent_handoff` event exists for it yet, emit the event then. A late-but-present journal record instead of a permanent gap.

## Relationship to other issues

Sibling to #955 (journal-mirror robustness — making per-dispatch signals survive timing/GC). Both stem from the same theme: *the journal is the durable record, so anything that should be durable must reach the journal regardless of write/complete ordering or GC timing.*

## Notes
- Severity: low — recoverable when work is committed; the failure mode is "durable HANDOFF record missing," not "work lost."
- The gate-side fix doubles as enforcement of an already-documented orchestrator rule.

michael-wojcik:
## Summary

Hardens the session journal as the GC-immune durable record by closing two ways durable signals are silently lost on long, high-dispatch sessions. Both issues share one root cause: the task store GCs older tasks faster than consolidation/wrap-up runs, and the journal is the GC-immune fallback — but some signals never reach it.

Closes #956. Closes #955.

## What landed

**#956 — `agent_handoff` write-after-completion race (`24ba784`)**
A HANDOFF-bearing teammate task completed *before* its `metadata.handoff` is written never re-emits `agent_handoff` (both emit paths gate on handoff presence *at completion time*). Two defenses:
- **Prevention** — a new **WARN-only** PreToolUse hook `handoff_ordering_gate.py` (`matcher:"TaskUpdate"`) that advises the lead when completing a HANDOFF-expecting teammate task with no handoff yet. Never denies (`additionalContext`, exit 0, fail-open on every path including module-load); the HANDOFF-expecting predicate reuses `is_self_complete_exempt` + the hoisted `is_teachback_subject` and keys on `is_lead`.
- **Guarantee** — a write-time **backstop** in `task_lifecycle_gate.py` that re-emits `agent_handoff` when the handoff lands on an already-completed task, idempotent via the shared `O_EXCL` marker (so `agent_handoff` now has three emit paths sharing one marker = exactly-once).
- Refactor: hoist `is_teachback_subject` to `shared/task_utils.py` so the new PreToolUse hook reuses it without importing the PostToolUse gate (clean one-way dependency direction).

**#955 — GC-immune per-dispatch variety + teachback-ack mirror (`b92c71d`)**
Per-dispatch variety (Task-B `metadata.variety`) and teachback acknowledgment (Task-A `variety_acknowledgment`) lived only in the GC-vulnerable task store, so wrap-up Q5/Q6 read false-empty after the reaper runs.
- Two new registered journal events `dispatch_variety` + `teachback_ack`, emitted hook-side from the lead-side PostToolUse gate (both `is_lead`-gated, fail-open).
- `wrap-up.md` Q5/Q6 redirected to read the journal first (GC-immune), falling back to the task store only when the journal yields nothing (exclusive-or, no double-count).

**Test hardening (`ac73683`)** — adversarial malformed-handoff coverage for the backstop (non-dict/empty handoff → zero emits, no O_EXCL poisoning).

## Why it's correct

- **D4 (the load-bearing assumption)** — that a metadata-only `TaskUpdate` reaches the PostToolUse gate — was confirmed at runtime via a reachability canary, then proven **non-vacuous by mutation** (neuter backstop → canary still passes [independence]; neuter the gate → canary fails [reachability]).
- The one fix-defeating risk (`dispatch_variety`'s `task_id` source) was **sealed against 16 real `TaskCreate` frames** from a live session transcript — all carry `tool_response.task.id`; the fix is not defeated.
- **Dual-mode**: every mode-divergent path keys on the structural `is_lead` discriminator (never a mode flag); the 13-row both-modes matrix is covered as a standing merge gate.
- Full suite: **8999 passed, 0 failed, 0 errors**.

## Scope / risk

Hooks-only; no external boundaries, dependencies, or credentials touched. Version bump: **PATCH 4.4.21 → 4.4.22** (additive hook + journal events + bug fixes).


michael-wojcik:
## Review complete — 4/4 SHIP, 0 blocking

Blind 4-reviewer panel (architecture, test, devops, security), all independent verdicts **SHIP**, 0 blocking findings. Highlights:

- **Marker idempotency** independently verified from two lenses (architecture + devops): the `O_EXCL` occupant key is `sha256(owner\0subject)` (not the PYTHONHASHSEED-salted builtin `hash()`), so the backstop re-emit is exactly-once across all 3 emit paths.
- **D4 non-vacuity** re-proven by mutation in both directions (neuter backstop → reachability canary still passes [independence]; invert the gate → canary fails [reachability]).
- **Security**: 24 live adversarial frames confirm the WARN gate never denies; path-traversal closed (`../../etc/passwd` → `etcpasswd`); and journal data exposure *narrows* — the `*_rationale`/`concern` free-text already lived in the `0o644` world-readable task store and now mirrors to the `0o600` owner-only journal.

## Post-review fold-ins (commit applying review feedback)

- **Trimmed `dispatch_variety`** to the spec shape `{4 dims + total}`, projecting via a `DISPATCH_VARIETY_KEYS` constant derived from the canonical `_VARIETY_DIMENSIONS` (so dimension names are never duplicated — neutralizing the drift surface reviewers flagged).
- Doc-comments: docstring rationale sharpening, the Q5 numerator/denominator coupling, the append-only no-dedup design, the backstop fail-open residual, and the exempt-set re-audit trigger.
- Added a `main()`-through-real-context advisory test (exercises the on-disk session-context resolution the fixture shortcuts past).

One future item — a real-tmux end-to-end seal of the backstop reach — is deferred to a follow-up issue (it can't run until this is the installed plugin). Full suite: **9000 passed / 0 failed / 0 errors**.

Issue #924, opened 2026-06-07
Codify post-merge live-probe as a standing gate for hook-infra PRs (+ reusable runbook)


michael-wojcik:
## Context

PR #920's #903 missed-wake surfacer shipped **INERT** — it never fired in live operation (`get_task_list` task-dir bug, #923). It was caught ONLY by an ad-hoc post-merge live-probe, **after** passing 2 full review rounds (10 reviewer-passes) + a verify-only re-review + 50 unit tests + an architect design-coherence verify. Static review and a mocked unit suite structurally cannot certify that a hook *fires* in real operation.

This recurs: prior instance of the "hook registered but not firing effectively" class is #551 (pact-memory `3d98819a`).

## Problem

The post-merge live-probe is currently an ad-hoc, **non-binding** wrap-up note. Per PACT's "collapses policy and implementation — there is no soft discipline layer" principle, a calibration memory (`6824953c`) will NOT change behavior. Without codification, the next hook feature can ship inert the same way — and be closed claiming resolution (as #903 was).

## Proposed

1. **Live-probe gate for hook-infra PRs.** In `wrap-up.md` (and/or `peer-review.md`), add a binding step: a PR that registers or modifies a hook MUST pass a post-merge live-probe — confirm the hook actually fires in a real session (**both teammateModes**) — before the cycle closes and before the originating issue is closed.
2. **Reusable live-probe runbook/skill.** Codify the probe pattern so it isn't re-invented each time: arm the precondition with back-dated/synthetic state → trigger the *real* hook event → verify the observable effect + the journal. The #903 probe (arm a stale `awaiting_lead_completion` task with a back-dated `since` → real `UserPromptSubmit` → confirm `additionalContext` + the `missed_wake` event) is the template.
3. **Non-mocked integration test for seam-dependent hooks.** Require that any hook whose value depends on an integration seam (task-dir resolution, env, real journal) has at least one NON-mocked integration test exercising that seam. The #903 suite mocked `get_task_list`, so the broken seam was the one every test stubbed (mock-hid-the-seam).

## Why a gate, not a note

The #903 inert-ship is direct evidence that a non-binding note is insufficient — the feature passed every static check and still shipped dead. Codification (a workflow step + a runbook + a test requirement) is PACT's only enforcement layer.

## References

- Motivating case: #903 (reopened) / #923 (the `get_task_list` team-dir fix)
[...]

PR #996, opened 2026-06-19
fix(hooks): live-probe gate fires under subdir project dir + surfaces WARN to agents (v4.4.32)

### LINKED ISSUE #932: locus-b live-probe gate (v4.4.14): two efficacy defects surfaced by the #924 tmux dogfood
## Context

The locus-b live-probe gate (`pact-plugin/hooks/live_probe_gate.py`, shipped in **v4.4.14** via #924) was dogfooded in its **TMUX / separate-process cell** per `pact-plugin/tests/runbooks/924-locus-b-dogfood-probe.md`, in a genuine separate-process session (3 distinct `session_id`s, separate iTerm2 panes — `session_id ≠ leadSessionId`, the §MODE MATRIX discriminator).

**The cell did NOT pass.** Two defects surfaced that the green static suite (`test_live_probe_gate_dogfood.py`) cannot catch — it certifies the decision *logic*, not real-process *resolution* or *surfacing*. Both are independently reproducible.

---

## Finding A — gate self-disables under a subdir `CLAUDE_PROJECT_DIR` (mode-divergence)

The gate was certified **mode-orthogonal** (`CLAUDE_PROJECT_DIR`-driven, config-dir-agnostic). It is not: it silently fail-safes from a process whose `CLAUDE_PROJECT_DIR` is the `pact-plugin` **subdir** (the platform sets a teammate's `CLAUDE_PROJECT_DIR` to its cwd = the subdir).

**Root cause** (`live_probe_gate.py`):

```python
def _resolve_repo_root() -> Optional[Path]:
    project_dir = os.environ.get("CLAUDE_PROJECT_DIR")
    if project_dir:
        return Path(project_dir)          # <-- returned UNCONDITIONALLY, no marker check
    # ... git rev-parse --git-common-dir parent (the correct repo root) ...

def _plugin_marker(root: Path) -> Optional[dict]:
    marker = root / "pact-plugin" / ".claude-plugin" / "plugin.json"   # assumes root == repo root
    ...
```

When `CLAUDE_PROJECT_DIR` = `…/PACT-prompt/pact-plugin`, `_plugin_marker` looks for `…/pact-plugin/pact-plugin/.claude-plugin/plugin.json` (**double `pact-plugin`**) → not found → marker `None` → fail-safe **silent**. The git-common-dir fallback (→ repo root, marker FOUND) is **shadowed** by the unconditional early return.

**Deterministic resolver matrix:**

| `CLAUDE_PROJECT_DIR` | marker | gate |
|----|----|----|
| repo root | FOUND v4.4.14 | fires |
| `pact-plugin` subdir | None | **silent (self-disabled)** |
| unset | FOUND (via git-common-dir) | fires |

**Blast radius (narrow):** canonical *lead-driven* merges run from the repo root (`CLAUDE_PROJECT_DIR` = repo root → marker resolves → gate fires). The silent self-disable bites only a merge/close driven from a subdir-cwd process.

**Proposed fix:** in `_resolve_repo_root` (or `_plugin_marker`), **validate the marker exists** at the `CLAUDE_PROJECT_DIR` root before trusting it (`_plugin_marker(Path(project_dir)) is not None`); if absent, fall through to the existing git-common-dir-parent path (and finally cwd). One guarded branch — preserves current behavior when `CLAUDE_PROJECT_DIR` is the repo root, and is robust to a subdir `CLAUDE_PROJECT_DIR`.

**Test requirements (fix A):** add an L1 test asserting a subdir-`CLAUDE_PROJECT_DIR` resolves to the repo root (marker FOUND), plus a **both-`CLAUDE_PROJECT_DIR` matrix** (repo-root vs subdir) per the dual-mode pin — mode-divergent resolution must be proven across both topologies or the orthogonality regresses silently again. Confirm the existing repo-root and unset cases stay green.

---

## Finding B — §2c surfacing is NEGATIVE for an agent operator

Forced past Finding A with a **temporary, reverted, md5-verified** one-line patch (`if False and project_dir:` → force the git-fallback = the *fixed* resolution; touches only root resolution, not the emit), the gate reaches `terminal=EMIT_WARN` (root=repo, marker 4.4.14, primary, satisfied=False) — but the `[live-probe-gate]` **stderr-on-exit-0 advisory never reaches the agent**, across two EMIT_WARN triggers.

**Positive control:** `gh pr merge --help` → `merge_guard_pre` **exit-2 deny DID surface** as `<error>`. ⇒ PreToolUse **exit-0 stderr is not fed to the model; exit-2 is.**

For **agent-driven merges (the dominant PACT case) the WARN is effectively invisible**, defeating the gate's purpose (remind the operator to run the live-probe before closing the issue). Human-terminal visibility is unverifiable from the agent vantage and is further undermined by the gate's `suppressOutput: true` stdout.

**Proposed fix:** switch `_emit_warn` to the isolated `hookSpecificOutput` informational fallback (additionalContext-style), which surfaces to the agent. The runbook §2c already anticipates this one-function swap; the gate already isolates it to one function.

---

## What DID pass (for the record)

- **WARN→SILENT mechanism** — a synthetic satisfied `4.4.14 | tmux PASS 2/2 · in-process PASS 2/2` row flipped `_has_satisfied_row` False→True → `terminal=silent:satisfied` (cross-row aggregation recognized the per-mode template). Row reverted (throwaway).
- **merge_guard interaction** — PreToolUse hooks do **not** short-circuit on a `merge_guard` deny (live_probe still runs); `gh pr close --help` (no `--delete-branch`) is the clean isolating trigger (merge_guard allows it, live_probe matches `_GH_PR_CLOSE_RE`).
- **Non-blocking posture** — the gate itself never denies (exit 0 always).
- **Separate-process mode** — 3 distinct session_ids, all `backendType=iterm2`.

---

## Acceptance / re-probe gate

Both fixes landed, **then a clean tmux re-probe** per `924-locus-b-dogfood-probe.md`, before #924 can close. **#924 stays OPEN.**

**Provenance:** surfaced by the #924 dogfood (tmux cell), 2026-06-10. Observed session_ids: lead `54ad4a1d…`, secretary `ee689856…`, devops `e9edfc32…` (3 distinct ⇒ separate-process / iterm2).

michael-wojcik:
## Summary

Fixes the two efficacy defects in the live-probe gate (`pact-plugin/hooks/live_probe_gate.py`) surfaced by the #924 tmux dogfood. **Implements the #932 code fixes (Finding A + Finding B); closure of #932 and #924 is pending the tmux re-probe** per `pact-plugin/tests/runbooks/924-locus-b-dogfood-probe.md` — that requires a real tmux teammate-mode session and is out of scope for this in-process change.

## What landed

**Finding A — gate self-disabled under a subdir `CLAUDE_PROJECT_DIR`.**
`_resolve_repo_root` returned `CLAUDE_PROJECT_DIR` unconditionally, so a process whose `CLAUDE_PROJECT_DIR` is the `pact-plugin` subdir resolved a doubled `pact-plugin/pact-plugin` marker path, found no plugin marker, and silently self-disabled — shadowing the correct git-common-dir fallback. The early return is now guarded on the plugin marker actually resolving at that root; otherwise it falls through to the existing git-common-dir-parent path, then cwd. One guarded branch — the **repo-root and unset cases are behaviorally unchanged**.

**Finding B — WARN advisory invisible to agent operators.**
`_emit_warn` printed its advisory to stderr on exit 0, which PreToolUse does not surface to an agent (only an exit-2 deny's stderr surfaces). The advisory now emits on stdout as `hookSpecificOutput.additionalContext` (`hookEventName: "PreToolUse"`), mirroring the verified sibling `task_claim_gate.py`. **Advisory prose is byte-unchanged and the gate still always exits 0 (non-blocking).** No `suppressOutput` co-emit on the WARN path (it would re-hide the advisory).

## Tests

- Comprehensive resolver matrix (repo-root / subdir / unset) + edge cases (marker-bearing dir trusted regardless of depth; cwd third-tier fallback) + an **end-to-end `main()`-WARNs-under-subdir** integration test (proves the gate *reconnects*, not just the resolver unit).
- Finding B: surfacing-shape assertions, advisory-prose byte-equivalence, and the silent-vs-WARN `suppressOutput` asymmetry negative control.
- **Non-vacuity** established by a source-only revert with clean per-finding failure attribution.
- Full hook suite: **9247 passed, 0 failed, 0 errors** (rtk proxy, explicit errors-token scan).

## Why

The gate exists to remind an operator to run the post-merge live-probe before closing a hook-infra issue. Both defects let it silently no-op for the dominant agent-driven-merge case — defeating its purpose. These fixes restore the intended advisory behavior.

Version: PATCH 4.4.31 → 4.4.32.

## Not closing yet

This PR does **not** close #932 or #924. Their closure is gated on a clean tmux re-probe per `924-locus-b-dogfood-probe.md` (which logs a satisfied both-mode row) — a real tmux teammate-mode session, sequenced post-merge.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.4.31",
+      "version": "4.4.32",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -605,7 +605,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.4.31/      # Plugin version
+│               └── 4.4.32/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.4.31",
+  "version": "4.4.32",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.4.31
+> **Version**: 4.4.32
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/hooks/live_probe_gate.py
+++ b/pact-plugin/hooks/live_probe_gate.py
@@ -204,8 +204,19 @@ def _resolve_repo_root() -> Optional[Path]:
         # be absent -> fall THROUGH to the git-common-dir-parent path (the true
         # repo root), then cwd. Preserves the repo-root case (marker found ->
         # trust it) and the unset case (this branch is skipped).
+        #
+        # The marker read here is repeated by main()'s own _plugin_marker(root)
+        # call after resolution — an intentional, accepted second read: two tiny
+        # file reads on a once-per-merge advisory hook, not worth refactoring the
+        # resolver's signature (e.g. returning the parsed marker) and taking any
+        # regression risk on this load-bearing identity resolver.
         return Path(project_dir)
     try:
+        # git-common-dir parent intentionally resolves to the MAIN checkout (not
+        # the worktree): the plugin marker is git-TRACKED, so it lives at the main
+        # root in every checkout, making it the stable IDENTITY anchor. The diff
+        # (Component C) is computed separately in the process CWD, so the current
+        # worktree's branch changes are still seen correctly.
         result = subprocess.run(
             ["git", "rev-parse", "--git-common-dir"],
             capture_output=True, text=True, timeout=5,
@@ -358,6 +369,13 @@ def _emit_warn(version: str, seam_hooks: frozenset[str]) -> NoReturn:
         f"closing the originating issue. Seam hooks touched: {seam}. "
         f"(Advisory — WARN only.)"
     )
+    # The WARN now rests SOLELY on this additionalContext channel — there is no
+    # secondary backstop (no stderr echo, no journal write). That is acceptable-
+    # by-design: this is a NON-BLOCKING advisory, so a missed reminder degrades
+    # to the (pre-existing) procedural CLAUDE.md pin, never to a blocked merge or
+    # a false "checked & clear" signal. The channel's runtime delivery is the
+    # subject of the post-merge tmux re-probe (924-locus-b-dogfood-probe.md),
+    # which confirms additionalContext actually surfaces before #924 closes.
     print(json.dumps({
         "hookSpecificOutput": {
             "hookEventName": "PreToolUse",  # MUST be the literal event name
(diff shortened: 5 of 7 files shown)


michael-wojcik:
Superseded by #997. The hook-audit this session confirmed `live_probe_gate.py` is maintainer-internal dev-tooling that leaked into the shipped plugin (1 of 25 hooks — the other 24 are genuinely consumer-facing). The right fix is to REMOVE the gate from the shipped plugin and replace the runtime-verification need with non-mocked seam-integration-tests — not the #932 efficacy fix this PR contained (which would have polished, and via Finding B made more visible, a hook we're deleting). Closing in favor of the removal tracked in #997.

Issue #1053, opened 2026-06-27
merge-guard: is_dangerous_command under-block — python -c subprocess.run list-form evades DANGEROUS_PATTERNS substring scan


michael-wojcik:
## Problem

`merge_guard_pre.py`'s `is_dangerous_command` detection is a regex substring scan over the command string (`DANGEROUS_PATTERNS` + `.search()`). A real destructive op invoked via the Python `subprocess.run` **list form** is NOT detected, because the list splits the op across elements and there is no contiguous `gh pr merge` substring to match:

```
python3 -c "import subprocess; subprocess.run(['gh','pr','merge','5'])"
```
→ `is_dangerous_command` = **False** (under-block: a real merge escapes detection).

By contrast the string / `os.system` form IS caught (contiguous literal):
```
python3 -c "import os; os.system('gh pr merge 5')"   # → blocked (correct)
```

## Class / scope

PRE-EXISTING limitation of the same **obfuscation-evasion class** already acknowledged in-code (the GraphQL/alias gaps near `merge_guard_pre.py` ~L183-185) and related to the `>32-flag-token` residual (#1001). The substring scan models contiguous command literals, not arbitrary programmatic reconstruction of the op tokens.

**Threat-model note**: the guard's threat model is honestly-authored command text by a trusted operator (preventing accidental/unapproved destructive ops), not an adversary deliberately reconstructing the op. The list-form is an unusual way to invoke a merge → **low–medium priority**.

## Provenance

Surfaced during the #1037 over-block-suppressor work and independently confirmed by an adversarial re-probe; explicitly kept OUT of #1037 scope (which is an over-block correction, not a detection-coverage expansion). The #1037 suppressor does NOT introduce or affect this gap.

## Possible direction (for discussion)

Detection-coverage expansion for programmatic op reconstruction (list-form, alias, GraphQL) is a larger change to the detection model — likely a coordinated effort across the known-gap class rather than a one-off. Must preserve INV-D2 (over-block acceptable, under-block never) and not regress the #1042/#1031/#1032 binding.

michael-wojcik:
[...]

Issue #1042, opened 2026-06-26
security: `--admin` and other privileged flags ride past merge-guard auth binding


michael-wojcik:
Discovered during PR #1040 / v4.4.43 (merge-guard auth-symmetry, #1031/#1032).

## Problem
The merge-guard binding model is `(operation, target)` — e.g. `(merge, 1040)`. The extractor drops dash-flags when finding the PR positional (`merge_guard_common.py`, ~L303). So an approved `gh pr merge 1040` and an executed `gh pr merge 1040 --admin` bind to the **same** `(merge, 1040)` — the `--admin` flag (which bypasses branch protection) rides past the guard undetected.

Same integrity-of-binding class as #1032 (approved-command == executed-command), but for privileged **flags** rather than op/target.

## Proposed direction
Extend the binding to a small denylist of consequential flags (`--admin`, force variants, possibly `--delete-branch`): such a flag must be present in the approved option to be permitted in the executed command, else refuse. Touches `merge_guard_common` + both hooks + tests + security/architect re-review.

## Related
multiSelect bundles are currently refused as a mint source (security-safe); revisit if privileged-flag approval ever needs multiSelect.

Severity: security hardening (narrow privilege-escalation-past-approval for a user who already has admin).

michael-wojcik:
## Post-install live-probe: PASS — closing per gated-closure discipline

The post-install live-probe gating this issue's closure is complete: **dual-independent convergent PASS** against the INSTALLED v4.4.44 merge-guard (`hooks/shared/merge_guard_common.py` + the real `merge_guard_pre.py` / `merge_guard_post.py` entry points), where installed v4.4.44 == merge commit `b01639d5` (the #1042 fix), lead-verified on disk before probing.

### Method
Two independent verifiers under an **independence guard** (each locked its full verdict before reading the other; reconciliation routed through the lead; honesty-disclosed all sources):
- **security-engineer** (primary probe): crafted-stdin driven through the **real hooks.json PreToolUse/PostToolUse entry points** as subprocesses, isolated `CLAUDE_CONFIG_DIR` (zero `~/.claude` contamination). Decision-only — no real destructive op executed.
- **test-engineer** (independent faithfulness verifier): deterministic-token READ arm + **genuine post-hook MINT seam** (`minted=True` per case), authored in an isolated `te_indep/` dir.

### Results — all cases, both arms (read-arm `pre`; mint→read seam `post→pre`)
| Case | Execute | Verdict |
[...]

Issue #1037, opened 2026-06-26
merge-guard: is_dangerous_command false-positive — `gh issue comment` body quoting a dangerous command is rejected (substring-scan brittleness)


michael-wojcik:
> ## ⚠️ STATUS UPDATE (2026-06-28) — general fix deferred; over-block accepted as safe
>
> Resolving this over-block via a **general** benign-arg-literal suppressor proved **intractable** for this SACROSANCT control. A HYBRID (regex-based, per-segment, verb-class-aware) suppressor was built and passed an 80-probe adversarial security re-probe + a GREEN auditor + 44 tests + a fully green suite — yet **two independent clean-room blind rounds each found NEW critical UNDER-blocks it introduced** (escaped-quote grammar; process-substitution `<(`, then `$(`/backtick; single `&`).
>
> **Root cause:** a regex cannot model bash grammar. In bash an "argument" can still *execute* (via `$()`, backtick, `<(…)`, `${ …; }`, `&`, `eval`, `bash -c`), so "the dangerous literal is only a quoted arg" is not a safe conclusion without a real parser — every regex patch closed the found case and missed adjacent siblings (whack-a-mole). The over-block is **safe** (over-block, never under-block); the under-block risk of a general fix is not worth it.
>
> **Disposition:**
> 1. **Partial safe fix** (separate PR): extend the proven, anchored gh carrier-strip to `gh issue/pr comment` (their `--body` is a non-executing API value, like create/edit). Fixes the gh-comment FP only.
> 2. The general over-block (`grep` / `git commit -m` / `pact_memory search` quoting a dangerous literal) is an **accepted safe limitation** — workaround: rephrase to avoid the literal.
> 3. A **sound general fix** needs a real bash parser (e.g. bashlex) with fail-closed-on-parse-failure — a larger, lower-priority project, deferred. This issue tracks it.
>
> **Sibling pre-existing UNDER-blocks** surfaced by the blind review (real security gaps, distinct from this over-block): #1053, #1055, #1056.
>
> ---
> *(original report below)*

## Problem

`merge_guard_pre.py`'s dangerous-command detection (`is_dangerous_command` / `DANGEROUS_PATTERNS`) scans the command *string* for dangerous-op literals (`git push --force`, `git branch -D`, `gh pr close --delete-branch`, etc.). To avoid false-positives when such a literal appears only as **non-executing quoted prose**, there is a `_strip_non_executable_content` pass that strips quoted carrier arguments before the scan.

[...]

PR #1054, opened 2026-06-28
fix(merge-guard): HYBRID benign-arg-literal suppressor for is_dangerous_command (#1037)

michael-wojcik:
## Summary

Resolves the #1037 `is_dangerous_command` substring-scan false-positive in the SACROSANCT merge-guard: benign command lines that merely **quote** a dangerous literal (a `gh issue/pr comment --body`, `grep`, `git commit -am`, `pact_memory search` argument) were over-blocked. This adds a HYBRID benign-arg-literal suppressor that runs **before** the unchanged `DANGEROUS_PATTERNS` scan, blanking provably-inert quoted literals of shell-inert-arg verbs only — never weakening detection.

**Completes ACs of #1037; closure pending post-install live-probe** (observable detection behavior changes — a previously-blocked benign line is now allowed).

## What landed (squashed from 4 commits)

- HYBRID suppressor `_suppress_benign_arg_literals` as the final step of `_strip_non_executable_content`, inherited by **both** call sites (`is_dangerous_command` + `is_compound_destructive_command`). Positive verb-whitelist (echo, grep, pact_memory search, gh issue/pr comment, git commit), fail-closed default.
- Quote-aware segment split (masks balanced quoted regions before `_COMPOUND_OPS_RE` separator-finding, reusing the #1000 `_gh_carrier_span` idiom; one segment SSOT).
- Placeholder-fail-closed defensive margin (`_segment_is_suppressible`).
- 42 comprehensive tests (measured non-vacuity; A/B suppressor-restraint-vs-neutral canary split; FP↔FN symmetry).
- Version bump → 4.4.45 (PATCH; over-block correctness, no user-facing capability change).

## SACROSANCT safety (INV-D2: over-block OK, under-block NEVER)

Three nested fail-closed gates keep it under-block-safe by construction (a suppressor bug reverts to over-block, never under-block):
1. **whole-command indirection** — any pipe-to-shell / process-sub / eval / command-substitution on the ORIGINAL command → suppress nothing;
2. **per-segment leading-verb whitelist** — a chained/wrapper-prefixed dangerous op stays caught;
3. **per-literal balanced-quote** — `$()`-containing double-quotes preserved; arg-executors (`python -c`, `eval`, `sh -c`) stay over-blocked by construction.

## Verification

- **Independent adversarial security re-probe = SAFE** (~80 probes on the final code, 0 under-blocks; mask/blank quote-pairing desync proven over-block-only; `$()` across all 3 quote contexts correct).
- **#1042/#1031/#1032 regression gate green** — no REFUSE→AUTHORIZE flip in the privileged-flag binding matrix (binding/token layer structurally untouched; this PR touches only `merge_guard_pre.py` + tests).
- **Full suite**: 9829 passed / 15 skipped / 0 failed / 0 errors (3.13.7 + 3.14.5).

## Known accepted residual (INV-D2-safe over-block)

The placeholder-guard introduces a narrow new over-block: a benign carrier whose quoted literal contains the substring `STRIPPED` is over-blocked (documented in-code). Over-block, not under-block — acceptable.

## Out of scope (tracked separately)

- #1053: pre-existing `subprocess.run` list-form under-block (DANGEROUS_PATTERNS substring gap) — not introduced or affected by this PR.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.4.44",
+      "version": "4.4.45",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -605,7 +605,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.4.44/      # Plugin version
+│               └── 4.4.45/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.4.44",
+  "version": "4.4.45",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.4.44
+> **Version**: 4.4.45
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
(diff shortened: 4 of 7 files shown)


michael-wojcik:
Superseded — closing in favor of #1058 (shipped as v4.4.45).

This PR's general benign-arg-literal suppressor was **abandoned**: two independent clean-room blind rounds each found NEW critical under-blocks it introduced, because a regex cannot model bash grammar (an "argument" can still execute via command-substitution, process-substitution, function-substitution, or a background separator). Patching each found case kept missing its adjacent siblings.

The **gh-comment** subset of #1037 is now resolved by the narrow, flag-anchored carrier-strip in #1058. The general over-block (grep / git commit -m / pact_memory search quoting a dangerous literal) remains an accepted safe limitation — a sound general fix would require a real bash parser, deferred and tracked in #1037 (re-scoped).

Pre-existing detection under-blocks surfaced by the blind review (independent of this PR): #1053, #1055, #1056. Workflow-discipline follow-up: #1057.

Issue #1088, opened 2026-07-02
Dangling #meta-block anchor: 3 consumers reference a slug with no host heading


michael-wojcik:
## What

The `#meta-block` anchor is referenced by 3 consumers across the plugin instruction files, but no host heading exists that yields that slug — the fragment is dangling (file-existence link tests pass; fragment resolution is untested).

Surfaced during the wake/idle ordering-hardening PREPARE sweep (anchor/cross-ref inventory over `pact-plugin/**/*.md`), where it was ruled out-of-scope drift (pre-existing, not co-located with the edited sections).

## Fix sketch

Either:
1. Add the missing host heading (e.g. `### META-BLOCK`) at the canonical META-BLOCK definition site in `protocols/pact-completion-authority.md` — NOTE: that file is a byte-compared extract; the heading must land in the same commit in its `pact-protocols.md` SSOT region (`scripts/verify-protocol-extracts.sh` gates it), or
2. Re-point the 3 referrers at an existing heading's slug.

Option 1 preferred: the referrers' intent is "the META-BLOCK rule", which deserves a stable anchor.

## Acceptance criteria

- [ ] `#meta-block` (or the replacement slug) resolves to a real heading in every referring file's target
- [ ] Extract/SSOT byte-parity maintained (verify script green)
- [ ] Consider a slug-derivation presence pin in `tests/test_wake_ordering_pinned.py` style so the anchor cannot silently dangle again

PR #1089, opened 2026-07-02
Agent Teams: wake/idle message-ordering hardening — disk-first re-read, boundary-drain, one-redundant-confirm (v4.4.52)

### LINKED ISSUE #1081: Harden Agent Teams instructions against wake/idle message-ordering races (crossed-wake at the teachback-acceptance seam)
## Class

Instructions-hardening (absorb a known platform behavior). NOT a platform bug report, and explicitly NOT fixable with a synchronous hook: `SendMessage` fires no Pre/PostToolUse event, inbox files are written asynchronously on delivery, and TeammateIdle notifications fire at turn-end regardless of what just landed in the recipient's inbox — synchronous wake/send detection is dead-by-construction (see the retirement of `completion_no_paired_send` in #897 and the deferred race-immune alarm direction in #903). The plugin's shipped instructions can, however, make agents *behave correctly under* the race. Today they don't.

## Observed failure signatures (live session evidence, 2026-07-02)

Orchestrating a 3-lane merge-guard batch (#1064/#1077/#1078) with teachback-gated dispatch produced **4 crossed-wake incidents across TWO distinct seams**, **4 stale counter-confirm messages**, and **1 mid-turn directive miss** (caught by lead-side verification before it reached a commit):

**Signature 1 — crossed-wake at ANY lead-wake-after-idle seam** (4 incidents: 3 at the teachback-acceptance seam, 1 at the stage-ready → lead-commit seam):

```
lead                            teammate
----                            --------
resolve teammate's wait
(accept teachback / commit):
  SendMessage(wake) ──────────► (in flight; inbox write is async)
  TaskUpdate(durable state)
                                turn ends → idle_notification fires
◄────────────────────────────── idle (postdates the wake-send)
reads teammate task: NOT advanced
sends ONE redundant confirm ──► processes inbox (wake + confirm)
                                proceeds
```

**Seam generalization (key finding)**: the fourth incident occurred at a *different* seam (stage-ready → lead-commit: the teammate idled on `awaiting_lead_commit`, the lead's commit-confirmation wake crossed its idle) — *predicted in an earlier revision of this issue, then observed*. The class occurs at EVERY boundary where a lead wake-SendMessage races a teammate's turn-end idle, i.e. every `intentional_wait` resolution. Mitigations must therefore be seam-AGNOSTIC: the disk-first re-read rule (below) must cover all `intentional_wait` reasons (`awaiting_lead_completion`, `awaiting_lead_commit`, teachback gates), not just teachback acceptance.

The lead cannot distinguish "idle because the wake hasn't landed yet" from "idle because the teammate stalled". Result per incident: 1 extra disk read + 1 redundant confirm + latency.

**Signature 2 — stale counter-confirm** (4 incidents): the teammate, processing the lead's redundant confirm after finishing the work, replies "crossed messages — task X is awaiting your validation/completion" based on a disk read that predates the lead's resolution writes. The assertion is already false on arrival; the lead must re-verify or ignore. Pure noise, and it can seed a politeness loop in both directions. All 4 incidents shared one shape: *a reply to a redundant confirm asserting state the lead had already resolved.*

**Signature 3 — mid-turn directive miss at a protocol boundary** (1 incident): a lead directive (a scope amendment) delivered while the teammate was mid-turn (mid-apply) surfaced only at the teammate's next idle boundary — but the teammate ran straight from applying to its stage-ready message without an intervening boundary, so the directive was never read. The teammate independently rediscovered *half* the amendment via its own sweep and mis-scoped the other half to a later phase ("FYI for TEST"). Caught only because the lead compared the reported stage scope against the outstanding directive before committing. (Post-incident, a "boundary-drain rule" was issued ad hoc to the remaining lanes and the next lane's stage-ready explicitly reported "boundary-drain done: inbox empty" — the rule is cheap and immediately usable.)

```
lead                            teammate (mid-turn, applying)
----                            --------
SendMessage(amendment) ───────► (lands in inbox; teammate never re-reads
                                 inbox before its boundary message)
                                ...applies, stages...
◄────────────────────────────── stage-ready (amendment absent)
compares scope vs directive:
MISMATCH → hold + re-instruct ─► re-applies, re-stages ✓
```

## Proposed instruction changes (no hooks, no protocol-shape changes)

1. **Teammate-side, `pact-agent-teams` SKILL (§On wake / Intentional Waiting)** — *disk-first re-read on wake, seam-agnostic*: on ANY wake while idling under ANY `intentional_wait` reason (`awaiting_lead_completion`, `awaiting_lead_commit`, teachback gate), FIRST re-read your gate/work tasks from disk (`status`, `blockedBy`, relevant metadata) and act on durable state; treat wake-message content as advisory confirmation. If durable state shows your wait resolved (gate completed / commit confirmation implied by task state) → proceed immediately, even if no wake message is visible yet. This makes the wake message redundant-by-design and ordering-immune at every seam.

2. **Teammate-side, same skill** — *counter-confirm suppression*: before sending any "crossed messages / already done / still awaiting" clarification, take a fresh disk read of the referenced task. If durable state already shows the situation resolved (e.g. the task you'd claim attention for is `completed`), send nothing. Durable state IS the reply.

3. **Lead-side, `pact-completion-authority` (or orchestrator persona §Completion Authority)** — codify *one-redundant-confirm-then-stop*: when an idle notification postdating your wake-send arrives and a durable read shows the teammate's wait unresolved, send exactly ONE redundant confirm naming the actionable state. Further idle ticks are not stalls; escalate to stall diagnosis only on task-file-mtime + sustained-silence evidence (per existing stall protocol). Today this discipline exists only as orchestrator-private memory; it should ship.

4. **Teammate-side, `pact-agent-teams` SKILL** — *boundary-drain rule*: immediately BEFORE composing any protocol-boundary message (teachback submit, stage-ready, HANDOFF notify, blocker report), re-check your inbox for directives that arrived mid-turn and reconcile them into the deliverable first; state the drain in the boundary message ("boundary-drain done"). A directive delivered while you are mid-turn is invisible until you look; the boundary message is your last chance to look before the lead acts on your word. (Field-tested in the same session: the next lane's stage-ready reported a clean drain.)

5. **Lead-side, orchestrator persona / `pact-completion-authority`** — *directive-reflection check*: before acting on a teammate's protocol-boundary message (especially stage-ready → commit), verify any directive you sent mid-turn is actually reflected in the deliverable (e.g. compare reported stage scope against outstanding amendments). Delivery ≠ processing.

## Acceptance criteria

- [ ] Instruction text added at the teammate-side surfaces (items 1, 2, 4) and the lead-side surfaces (items 3, 5) above.
- [ ] Item 1's rule is written seam-agnostically (all `intentional_wait` reasons), not teachback-specific.
- [ ] Both-teammateMode applicability stated (in-process AND tmux) — the race is delivery-ordering, not mode-specific.
- [ ] Explicit non-goal note: no hook-based detection (dead-by-construction; cite the async-inbox contract) — prevents a future "fix" reintroducing the #897 false-positive class.
- [ ] No change to the two-call atomic pair itself (SendMessage-first ordering stays load-bearing).

## Evidence tally

- 2026-07-02 session (team session-03047b41): **4× Signature 1** (3 at teachback-acceptance, 1 at stage-ready→lead-commit — the predicted second seam, confirming seam generality), **4× Signature 2** (uniform shape: replies to redundant confirms asserting already-resolved state), **1× Signature 3** (scope amendment missed mid-apply; lead directive-reflection check caught it pre-commit, one re-stage cycle to recover; boundary-drain rule field-validated on the following lane). 0 lost work. Tally final-checked at session wrap-up.

michael-wojcik:
## Summary

Hardens the Agent Teams instruction surfaces against wake/idle message-ordering races (crossed-wake idles, stale counter-confirms, mid-turn directive misses). Instruction-only — no hooks, no protocol-shape changes; synchronous wake/send detection remains dead-by-construction (unhookable SendMessage + async-on-delivery inbox writes), so the fix makes agents behave correctly *under* the race.

Closes #1081. Follow-up filed: #1088 (pre-existing dangling `#meta-block` anchor, out of scope).

## What landed (5 commits)

1. **`skills/pact-agent-teams/SKILL.md`** — teammate-side: seam-agnostic **On-Wake Disk-First Re-Read** (any `intentional_wait` reason + any crossed inbound directive, per the issue comment's generalized trigger), **Counter-Confirm Suppression** (complement, not substitute), **Boundary-Drain Rule** (mandatory literal inbox-file read before every protocol-boundary message, best-effort/fail-open, idempotent reconciliation, REQUIRED drain statement); fixes stale no-hook claims.
2. **`protocols/pact-completion-authority.md` + `protocols/pact-protocols.md` (byte-identical SSOT mirror) + `agents/pact-orchestrator.md`** — lead-side: **Crossed-Wake Idles: One Redundant Confirm, Then Stop** (incl. the behavioral no-hook non-goal note), **Directive-Reflection Check** (delivery ≠ processing); corrects the stale 4-point-rule summary down to the protocol's 3-point rule (retired token dropped) and stale no-consumer claims.
3. **`protocols/pact-agent-stall.md` + SSOT twin** — harmonizes "treat as stalled immediately" with the new rule: post-wake / live-`intentional_wait` idles are delivery-ordering artifacts, not stalls.
4. **`tests/test_wake_ordering_pinned.py`** — 53 presence pins (line-anchored headings, whitespace-normalized phrases robust to rider/wrap renderings, anchor-slug cross-ref + slug-derivation integrity checks, retired-token absence guards). Counter-tested against pre-hardening text: 45/45 revert-couplable pins flipped RED.
5. **Version bump** → 4.4.52 (PATCH: instruction refinement).

## Why

Live evidence from two orchestration sessions (6+ crossed-wake incidents across three seam types, 5 stale counter-confirms, 1 mid-turn directive miss). The rules were dogfooded during their own implementation: the coder's boundary-drain caught a crossed lead confirm mid-turn and reconciled it idempotently — first in-situ field validation.

## Verification

- All ACs verified against the shipped diff (seam-agnostic wording; both-teammateMode applicability; behavioral no-hook non-goal, zero issue/PR refs in LLM-loaded files; two-call atomic pair + 3-step ordering invariant byte-untouched).
- `verify-protocol-extracts.sh` 19/19 MATCH (mirrors byte-identical by construction — same Edit pair applied to both sides).
- Existing count pins intact ("Ordering invariant" == 3; wake-signal marker phrase 2/2/2/1/2).
- Full suite on the CI-faithful venv (python 3.13 + latest deps per `.github/workflows/tests.yml`): **10430 collected, 0 failed, 0 errors** (explicit errors-token scan).
- Concurrent auditor signal: GREEN (byte-parity, pin counts, verbatim fidelity, cleanliness — verified against `git diff` ground truth).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.4.51",
+      "version": "4.4.52",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -605,7 +605,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.4.51/      # Plugin version
+│               └── 4.4.52/      # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.4.51",
+  "version": "4.4.52",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.4.51
+> **Version**: 4.4.52
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/agents/pact-orchestrator.md
+++ b/pact-plugin/agents/pact-orchestrator.md
@@ -138,7 +138,7 @@ Certain conditions bypass normal orchestration and escalate directly to user:
 
 When waiting for teammates to complete their tasks, **do not narrate waiting** — saying "Waiting on X..." is a waste of your context window. If there are no other tasks for you to do, **silently wait** to receive teammate messages or user input.
 
-Idle notifications arrive as conversation turns. When a turn carries no actionable content — no blocker, no stage-ready, no question, no user input — emit no reply. Acknowledging every incoming turn is the reflex that produces narrate-the-wait noise. The next meaningful transition triggers the next meaningful reply.
+Idle notifications arrive as conversation turns. When a turn carries no actionable content — no blocker, no stage-ready, no question, no user input — emit no reply. Acknowledging every incoming turn is the reflex that produces narrate-the-wait noise. The next meaningful transition triggers the next meaningful reply. One protocol-defined exception: the single redundant confirm after a crossed wake (an idle notification postdating your wake-send) — see §12 Intentional Waiting.
 
 ---
 
@@ -486,7 +486,7 @@ Teammate self-completion carve-outs (predicate-witnessed): signal-tasks (`metada
 
 ### Teachback Review
 
-Each specialist dispatch is a Task A (TEACHBACK) + Task B (work) pair with `blockedBy=[A]` — see §11 for the canonical sequence. Teammate claims A, writes `metadata.teachback_submit` (5 canonical fields per [pact-teachback](../skills/pact-teachback/SKILL.md)), idles on `awaiting_lead_completion`. **Read-Trigger Precondition**: wait for teammate's wake-signal SendMessage BEFORE the raw JSON read — see [pact-completion-authority §Read-Trigger Precondition](../protocols/pact-completion-authority.md#read-trigger-precondition) for the 4-point rule (Monitor `INBOX_GREW` is alarm-clock not content marker; raw read MUST follow SendMessage receipt; mitigation for residual race). Then read the payload via raw JSON (TaskGet is metadata-blind), apply Validating Incoming Teachbacks below, then accept via the Acceptance two-call atomic pair above; acceptance auto-unblocks Task B. Do NOT mark Task B `completed` or `pending` yourself — the teammate claims on wake.
+Each specialist dispatch is a Task A (TEACHBACK) + Task B (work) pair with `blockedBy=[A]` — see §11 for the canonical sequence. Teammate claims A, writes `metadata.teachback_submit` (5 canonical fields per [pact-teachback](../skills/pact-teachback/SKILL.md)), idles on `awaiting_lead_completion`. **Read-Trigger Precondition**: wait for teammate's wake-signal SendMessage BEFORE the raw JSON read — see [pact-completion-authority §Read-Trigger Precondition](../protocols/pact-completion-authority.md#read-trigger-precondition) for the 3-point rule (wake-signal SendMessage is the content-arrival signal; raw read MUST follow SendMessage receipt; mitigation for residual race). Then read the payload via raw JSON (TaskGet is metadata-blind), apply Validating Incoming Teachbacks below, then accept via the Acceptance two-call atomic pair above; acceptance auto-unblocks Task B. Do NOT mark Task B `completed` or `pending` yourself — the teammate claims on wake.
 
 #### Validating Incoming Teachbacks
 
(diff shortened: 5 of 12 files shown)


michael-wojcik:
**Copilot review disposition** (both findings evaluated against HEAD):

1. **Inbox `summary` field** — resolved in ef655ffd (before this review was read): the field list now names `from`/`text`/`timestamp`/`type` with a platform-version hedge, grounded in an empirical 132-message survey. The remaining "summary" token in the template is the drain report's own `<one-line summary>` placeholder, not an inbox-schema claim.

2. **`_github_slug()` hyphen normalization** — declined. GitHub empirically does **not** collapse repeated hyphens or strip edge hyphens: an em-dash strips to empty, yielding `--` in the real anchor (verified previously in this repo via the `application/vnd.github.html` API). The current helper reproduces that behavior exactly (dropped em-dash leaves two spaces → `foo--bar`); the proposed normalization would make the helper diverge from real GitHub anchors. The helper is additionally run-validated against all four in-scope headings.

PR #1183, opened 2026-07-15
fix(dispatch): resolve resume-path stale-team split-brain permanently

v4lheru:
## Problem

After \`claude --resume\`/restart, the dispatch gate intermittently reports **\"no Task assigned to owner=...\"** even though \`TaskList\` clearly shows the tasks. This blocks work frequently (multiple times a day for heavy \`--resume\` users) and the 4.6.5 self-heal + stale-diagnosis do not reliably fix it.

## Root cause

One PACT session spawns **many** team dirs (every \`Agent\`/comPACT/orchestrate), all stamped with the lead's \`leadSessionId\`. \`_resolve_aligned_team_name\` identity-matches \`leadSessionId == session_id\` but tie-breaks by **sorting the directory name** and returning the alphabetically-first — which is unrelated to which team the harness actually routes tasks to. On-disk census confirmed the collision: 59 team dirs, with 14 sharing one \`leadSessionId\` and 12 sharing another. \`session_end\` never retired any of them, so the collision corpus grows every session. The 4.6.5 self-heal reconciles *toward* the (stale) identity-matched value, and the stale-diagnosis returns \`None\` on the same-id case — so neither saves the user.

## Fix (3 commits)

1. **\`0f5e94c1\` — resolver (P0, the fix).** Collect *all* genuine \`leadSessionId==session_id\` matches; on collision (len≥2) disambiguate via an identity-anchored, liveness-ordered ladder: **caller default anchor → own-canonical-name {session_id, session-<id8>, pact-<id8>} → task-store child-mtime recency → createdAt-max → smallest name** (parity with prior \`sorted()[0]\`). len==0/len==1 paths are byte-identical (zero regression); new disk reads degrade the sort key without raising (never-raises totality preserved). Identity anchors precede liveness so an active sub-team with a fresher task store can never steal the lead's own team.
2. **\`2045f857\` — session_end hygiene (P1).** Retire (\`leadSessionId=None\`) the ending session's **non-live** same-id siblings — skips the live team (case-insensitive) and foreign-id dirs, config-only atomic rewrite, **no rmtree**, separate pass from the registry prune, fail-closed (runs only when both session_id and team_name are known). Stops the collision corpus from growing.
3. **\`12d4e84e\` — regression suite.** 33 tests on top of the coder's 18 smoke tests: both-topologies integration through the real \`get_team_name\` gate (proves the split-brain is closed), strict tier precedence, all pinned invariants (empty-SSOT fail-closed, foreign-id never promoted, branch-2 config-less fallback, never-raises under malformed inputs, retirement safety), and a **counter-test-by-revert** confirming the tests fail without the fix.

## Verification

- Full suite: **11175 passed** (+29), 445 skipped, 1 pre-existing unrelated failure (\`test_memory_cli\`, confirmed identical on the pre-fix baseline).
- Independent concurrent audit: **GREEN** against git diff *and* live runtime.
- All 11 pinned invariants preserved; correctness target = the operator recovery helper \`pact-align.py\` steps 3+4.

Built via full PACT cycle (PREPARE → ARCHITECT → CODE+audit → TEST). Design docs in \`docs/preparation/\` and \`docs/architecture/\`.

### DIFF
--- a/pact-plugin/hooks/session_end.py
+++ b/pact-plugin/hooks/session_end.py
@@ -24,6 +24,7 @@ import os
 import re
 import shutil
 import sys
+import tempfile
 import time
 from pathlib import Path
 
@@ -788,6 +789,121 @@ def _prune_registry_dead_teams(
     return pruned
 
 
+def _atomic_write_config(target: Path, data: dict) -> None:
+    """Atomically rewrite ``target`` (a team ``config.json``) with ``data``.
+
+    Temp file in the SAME directory + ``os.rename`` (crash-safe atomic replace),
+    0o600 — mirrors ``pact_context.persist_context``. Raises on any failure so
+    the caller's per-dir guard counts the dir as a skip rather than a retirement
+    (a partial/failed write must never be recorded as success). The temp file is
+    cleaned up on failure. Not fail-safe by itself BY DESIGN — the caller owns
+    the never-raises boundary.
+    """
+    parent = target.parent
+    fd, tmp_path = tempfile.mkstemp(dir=str(parent), prefix=".config-", suffix=".tmp")
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            json.dump(data, f)
+        os.chmod(tmp_path, 0o600)
+        os.rename(tmp_path, str(target))
+    except Exception:
+        try:
+            os.unlink(tmp_path)
+        except OSError:
+            pass
+        raise
+
+
+def _retire_session_leadsessionid_claims(
+    current_session_id: str,
+    current_team_name: str,
+    teams_dir: Path | None = None,
+) -> int:
+    """Retire (``leadSessionId = None``) the ENDING session's claim on its
+    NON-LIVE sibling team dirs — the SessionEnd hygiene half of the
+    resume-path split-brain fix (issue: dispatch-gate stale-team split-brain).
+
+    A single PACT session stamps its ``leadSessionId`` onto MANY team dirs
+    (every Agent/comPACT/orchestrate mints one). Nothing retires those claims
+    today, so the collision corpus the resolver must disambiguate grows every
+    session. This pass mirrors ``pact-align.py`` step 4: when THIS session ends,
+    null out its ``leadSessionId`` on every sibling dir EXCEPT the live team —
+    collapsing the collision toward a single claimant for the next resume.
+
+    Predicate for retiring a dir ``D`` (ALL must hold):
+      * ``D`` is a real directory (symlinks skipped — lstat semantics — so a
+        planted link cannot redirect the rewrite outside the teams root);
+      * ``D/config.json`` exists and parses;
+      * ``data['leadSessionId'] == current_session_id`` — ONLY this session's
+        claims. A dir claimed by a DIFFERENT (foreign) session is left
+        untouched; an already-retired ``None`` claim is likewise ``!=`` our id
+        and skipped, making a re-run a no-op (idempotent);
+      * ``D.name.lower() != current_team_name.lower()`` — the LIVE team KEEPS
+        its claim (pact-align step 3 parity), so the collision collapses to
+        exactly one live claimant.
+
+    Action on a match: read-modify-write ``config.json`` setting
+    ``leadSessionId = None`` (JSON null), preserving all other fields, via an
+    atomic temp-file rename. NO ``rmtree`` — the platform owns dir teardown; we
+    retire the CLAIM, not the DIR (leaves ``session-*`` / random-named dirs in
+    place). A SEPARATE pass from ``_prune_registry_dead_teams`` (registry JSONL)
+    and ``cleanup_old_teams`` (TTL rmtree) — it touches only ``config.json``.
+
+    FAIL-CLOSED: without BOTH ``current_session_id`` AND ``current_team_name``
+    we cannot guarantee we would SKIP the live team, so we retire NOTHING (never
+    risk nulling the live team's own claim). Best-effort / never raises: a
+    per-dir ``try/except`` swallows unreadable/malformed configs and write races
+    so a retirement failure can never block session termination.
+
+    Args:
+        current_session_id: the ending session's id (the claim to retire).
+        current_team_name: the live team dir name to PRESERVE (skip).
+        teams_dir: the live-teams root. Defaults to ``~/.claude/teams``.
+
+    Returns:
+        Number of dirs whose ``leadSessionId`` was retired (0 on fail-closed
+        no-op, absent teams root, or nothing to retire).
+    """
+    # FAIL-CLOSED guard (belt-and-suspenders with the main() callsite check).
+    if not current_session_id or not current_team_name:
+        return 0
+    if teams_dir is None:
+        teams_dir = get_claude_config_dir() / "teams"
+    base = Path(teams_dir)
+    if not base.exists():
+        return 0
+
+    live_name = current_team_name.lower()
+    retired = 0
+    try:
+        for entry in base.iterdir():
+            try:
+                # Skip symlinks (lstat) and non-dirs; skip the LIVE team.
+                if entry.is_symlink() or not entry.is_dir():
+                    continue
+                if entry.name.lower() == live_name:
+                    continue
+                config_path = entry / "config.json"
+                data = json.loads(config_path.read_text(encoding="utf-8"))
+                # Only OUR session's claims — foreign ids and already-retired
+                # (None) claims are both != our id, so both are skipped.
+                if not isinstance(data, dict):
+                    continue
+                if data.get("leadSessionId") != current_session_id:
+                    continue
+                data["leadSessionId"] = None
+                _atomic_write_config(config_path, data)
+                retired += 1
+            except (OSError, json.JSONDecodeError, ValueError, TypeError,
+                    AttributeError):
+                # Unreadable / malformed / write race → skip this dir, keep
+                # going. A retirement failure must NEVER block termination.
+                continue
+    except OSError:
+        pass  # iterdir race on the teams root → best-effort, never raise
+    return retired
+
+
 def main():
     try:
         try:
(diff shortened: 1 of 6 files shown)


michael-wojcik:
## Review complete — four independent lenses, and the verdict splits per commit

Thanks for this. It got a full multi-agent review (architect / test-engineer / backend-coder / security-engineer), and the short version is: **the code is clean — three of four lenses found 0 Blocking. The disagreement is about P1's design and about the premise, not your craft.**

**Recommendation: split the PR. P0 is viable with two fixes. P1 should come out.**

---

### What you got right (worth saying explicitly)

- **Your test call sites all pass explicit `teams_dir=`.** The writer defaults to the real `~/.claude/teams` — one omitted kwarg would have rewritten a reviewer's live configs. You didn't make that mistake, and it's a non-trivial one to avoid.
- **The case-insensitive live-team skip is correct**, not a bug. macOS/Windows FS are case-insensitive, so a case-*sensitive* compare could fail to recognize the live team and retire it — the catastrophic direction. The residual false-negative is under-retirement = fail-safe. Right trade.
- **The write is genuinely atomic** (mkstemp same-dir + rename + 0o600 + unlink-on-failure). Cannot truncate. "Mirrors `persist_context`" is accurate.
- **Leaving the identity predicate untouched** is what prevents the ladder from reconciling toward a stale value — i.e. exactly the flaw you (correctly) indict 4.6.5 for. Your predicate placement is what saves it.
- **`leadSessionId: null` is not activated**: all four readers degrade fail-safe. A retired team becomes indistinguishable from an unreadable config, and every consumer's unreadable-branch is already the conservative one.
- **Your test numbers are honest, not wrong.** 11175+445+1 = 11606+15+0 = **11621 — identical total collection.** Your environment skips 430 tests that execute and pass here, and the `test_memory_cli` failure doesn't reproduce. Your green is a weaker run, not a false one.

---

### P0 (`0f5e94c1`) — viable. And it works.

We initially concluded the ladder was dead code. **We were wrong, and we proved ourselves wrong.** Cross-era resume, measured:

| | Pre-resume | Post-resume |
|---|---|---|
| `matches` | 1 | **2** |
| `_disambiguate_collision` | not invoked | **fires**, invocations=1 |
| Pick | legacy `pact-<id8>` | **`session-<id8>`** — the platform's real, task-store-live team, via tier-1 |
| Retirement | `retired=0` | collapses the legacy claim 2→1 — exactly its stated purpose |

`session_init`'s `persist_context` does an unconditional `os.rename` on every lead SessionStart, so the stale ctx name is replaced and tier-1 anchors on the live team. **On the one path we could demonstrate, your fix does the right thing.**

**Two fixes needed before merge:**

1. **Perf regression — 100% of consumers pay, 0% get the fix.** The collect loop dropped the old early-return, so every fresh hook process now full-scans `teams/`. Measured on a real 226-dir root: early-match **1.84ms** vs no-match **1.81ms** — ratio 1.02, no short-circuit. The docstring still claims *"on a MATCH the scan stops at the first matching dir"* (~0.45ms over ~21 dirs). **The change turned a true claim false.**
2. **Two provably vacuous tests** (one-word fixes). `test_tier1_anchor_beats_canonical_name` and `test_own_canonical_name_wins_when_default_absent` — both fixtures sort *before* the target, so the old name-sort returns the ladder's answer. They can't discriminate the fix. Pick fixture names where the old sort returns a *different* dir. Verified two-leg: rename → fails on reverted baseline, passes on the fix.

Also: PR body says "33 tests"; actual is **29**. "18 smoke tests" is exact (8+10). Total new: **+47**.

Minor: `_disambiguate_collision` copies the id8 regex while its docstring claims it "reuses generate_team_name's exact id8 sanitizer." Byte-identical today; duplicated logic that can silently diverge.

---

### P1 (`2045f857`) — please drop this one

This is the substantive objection, and it needs no data from your machine:

> **P1 fires at SessionEnd and treats "ending" as "dead." `claude --resume` reuses the session_id and is first-class in PACT. A session that ends is *dormant*, not dead — so P1 tombstones claims belonging to a session that may return.**

A PR whose entire purpose is fixing the resume path tombstones state on the assumption that sessions don't resume. That's unconditional and reachable on every machine.

Structurally: **P1 has no liveness probe at all.** Its "live team" notion is an identity compare against `get_team_name()` (`session_end.py:945`) — which is *P0's own resolver*, the one whose mis-resolution is the reported bug. When the resolver is right, P1 is unnecessary. When it's wrong — the exact case motivating this PR — P1 skips the wrong dir and **retires the right dir's claim**. Reproduced with a control: with P1, `self_heals=FALSE, locked_in=TRUE`; without P1, the same wrong pick stays correctable because both claims survive. Its guard catches *empty* skip-keys, never *wrong* ones, despite promising it will "never risk nulling the live team's own claim."

Net: **P1 is safe where the bug is absent and potentially permanently harmful where it's present** — the risk concentrates on exactly the machine that needs the fix. And the collision-corpus growth it targets isn't happening: UUID-form dirs here are n=208, oldest 27.7d, **zero past the 30-day TTL**.

---

### Two questions — only you can answer these

Your **symptom is real** and we're not dismissing it. But your **stated mechanism doesn't reproduce**: spawns add *members*, not dirs — verified across all three naming eras. A 2026-06-15 legacy team here holds **11 members in one `config.json`** under one `leadSessionId`. So the 14-way collision is unexplained by anything we can find in PACT's code or on our disk.

It is **not impossible**, though — three two-dir pairs exist here right now, held apart only by the fact that at most one dir per pair carries a `config.json`. That's *one `config.json` away* from your topology.

1. **`claude --version`**, your teammateMode, and whether you launch via CLI or Desktop.
2. **Do you have both a full-UUID and a `session-<id8>` team dir for the same session, *each with a `config.json`*?**

If the answer to (2) is yes, your premise is real and version/launcher-bounded — not wrong — and this becomes a very different conversation. If max `leadSessionId` multiplicity is 1 on your disk too, then whatever is breaking your `--resume` is something else and worth its own investigation (our leading unverified candidate: task-namespace divergence between write-time and read-time `get_team_name()`).

### If your environment does justify it — a ~4-line alternative

Tier-2's canonical scan alone, inside the same `len>=2` gate: drop `tasks_dir`, `_task_store_mtime`, `_safe_created_at` and every task-store read. Tiers 3–5 only fire when no dir is named after this session — at which point the naming invariant is already broken and mtime/createdAt are heuristics rather than resolution. Tier-1 is also strictly redundant with tier-2 on every reachable path.

v4lheru:
Thanks — this is an exceptional review. The empirical work (measuring the ladder actually fire cross-era, the two-leg test verification, the symlink-write repro) is genuinely appreciated, and I'm agreeing with the split. Point by point:

## Dropping P1 — agreed
You're right, and the argument holds independent of environment. P1's notion of "live team" is an identity compare against `get_team_name()` — the very resolver whose mis-resolution *is* the bug — so in the exact case this PR exists for, P1 retires the **right** dir's claim and locks in the wrong pick, destroying P0's self-heal. Treating SessionEnd as "dead" also breaks the `--resume`-reuses-`session_id` contract. It's out. The two commits were built independently-revertable (disjoint files, invariant sets 1–6 vs 7–8), so `2045f857` drops cleanly with zero rework to P0.

## P0 fixes — all landing
- **Docstring early-return claim:** correct — the collect loop dropped the short-circuit; the docstring goes stale-claim → accurate. (Your 1.02 ratio confirms it's a doc-accuracy bug, not a meaningful perf hit, but the false claim shouldn't ship.)
- **The two vacuous tests** (`test_tier1_anchor_beats_canonical_name`, `test_own_canonical_name_wins_when_default_absent`): good catch — fixtures where the old name-sort already returns the ladder's answer can't discriminate the fix. Re-seeding with names where the pre-fix sort differs, verified two-leg (fails on reverted baseline, passes on fix).
- **PR body 33 → 29** (18 smoke exact), plus the id8-regex duplication-vs-docstring nit.

## Your two questions — with disk data
`claude --version` **2.1.210**, teammateMode **default (in-process)**, launched via **CLI**.

Q2 (a full-UUID *and* a `session-<id8>` dir for one session): **no** — that's not the shape here. But the collision is real and reproducible on my disk, via a different mechanism than either of us guessed:

- **14 separate team dirs**, each with its own `config.json`, all stamped `leadSessionId=0be9512d`; another **12** sharing `64f7b112`.
- They're **random-named** (`agile-swinging-shore`, `composed-plotting-naur`, `wild-prancing-comet`…), most just `['team-lead']` — and some host real work (`wild-prancing-comet` = `['team-lead', 'router-fix-coder']`).
- Dir-shape census: **0 full-UUID, 10 `session-<id8>`, 50 random-named.**

So on 2.1.210 / in-process, the harness mints a **fresh random-named team per workflow**, all under the lead's `session_id` — that's the collision source, not spawns-adding-members-to-one-config. Version/launcher-bounded, exactly as you allowed for.

## Pushing back on the tier-2-only alternative
This is the one place I'd keep tiers 3–5. The simplification assumes the live team is always `session-<id8>`-named, so tier-2 catches it. On my disk the live/collision teams are **random-named**, and tier-2 only recognizes `{session_id, session-<id8>, pact-<id8>}` — so it finds no match, and **tier-3 (task-store recency) is precisely what disambiguates them.** Dropping tiers 3–5 fixes the `session-<id8>` case but leaves the random-named case — the majority here (50 vs 10) — unresolved. Task-store recency is load-bearing *for the environment that actually has the bug*.

## The bigger finding: it's two bugs — and the second reproduced live
Resuming **this very session** to write this reply, the blocker fired again — and it was **not** the collision. It was your **"task-namespace divergence between write-time and read-time `get_team_name()`"** candidate, precisely:

- The harness minted `session-287d689b` (its `config.json` `leadSessionId=287d689b…`, holding the freshly-created task).
- The plugin's persisted context + CLAUDE.md still said `session-df19e9cd` — which **had no `config.json` at all** — so the gate resolved a dead team and denied dispatch while the task sat in the live team.

**P0's ladder does not fix this** — there's nothing to disambiguate; the persisted `session_id` itself is stale and no dir claims it. It needs a **SessionStart-time write-back** re-keying the persisted team/session to the harness's live team (a one-command manual align already recovers it). So:

- **P0** fixes the *collision* mechanism (tier-3 needed for random-named teams).
- **Namespace-divergence** is a separate fix — I'll open it as its own issue/PR rather than smuggle it into this one.

Revised P0 (P1 dropped, four fixes applied) incoming. Thanks again — this made the fix materially better.

Issue #1265, opened 2026-07-26
Test suite writes to the live pact-memory DB, and the Working Memory sync cites IDs that only exist in a temp DB


michael-wojcik:
> ## ⚠️ REVISION 2026-07-26 — LEAK A CONFIRMED AND NARROWED; LEAK B IS ON A DIFFERENT ROUTE THAN THIS BODY DESCRIBES
>
> Six regime cells measured in a sandbox (throwaway `HOME`, scratch repos, pre/post integrity assertion on the real store). Corrections below; the original body is retained beneath.
>
> ### Leak A — CONFIRMED, and smaller than measured here
>
> **41 bare call sites, not 44.** The census above was line-oriented and could not see three multi-line calls passing `db_path` on a continuation line (`test_archive_pin.py:577`, `test_archive_pin_emit.py:557`, `test_archive_pin_resolver.py:329`). Commit-time `00:01:06Z` vs this issue's comment at `00:17:21Z` establishes they were **missed, not added**.
>
> **Exactly one site reaches a real save**: `test_archive_pin_emit.py:180`, parametrized ×2. Classification of all 51: SCOPED 10 / STUBBED 31 / SHORT_CIRCUIT 9 / **REACHING 1**. 1 site × 2 params × 15 runs = **30** — matching this body's measured 30 rows at `goal = "Archived pin: First Pin"`, at exactly two `project_id`s of 15 each. **The other 12 unscoped sites are blast radius, not current writes.**
>
> ⚠️ **13 of the 43 rows remain unattributed.** The 8 (goal: *a copy of a real pin, from a fixture*) and the 5 (`Archived pin: Alpha`) carry **different goals** and cannot come from `emit:180` in any regime — that site always emits "First Pin"; the regime moves only `project_id`. Had it produced 30 + 8, the First Pin count would be 38. It is 30. **Their producing sites are unidentified.**
>
> ⚠️ **A required `db_path` parameter does NOT prevent `db_path=None`** — it forces callers to *name* it. And `archive_pin.py:470` tests **truthiness**, so `db_path=""` routes to production. Any falsy-rejection guard **must be gated on `PYTEST_CURRENT_TEST`**: `main():906-909` passes `args.db_path = None` for the curator's normal production run, so an ungated guard breaks `archive_pin --index 0` — the command `/PACT:prune-memory` keys its refuse-or-proceed decision on.
>
> ### Leak B — the remedy in this body is a category error
>
[...]

Issue #1240, opened 2026-07-22
pact-memory: the Working Memory sync target resolves ambiently with no caller-specifiable override (test-suite symptom fixed in 0be93d08; root open)


michael-wojcik:
> ## ⚠️ REVISION 2026-07-26 — SYMPTOM FIXED, ROOT OPEN. RETITLED.
>
> **The fix this body prescribes has shipped.** `tests/test_memory_cli.py:104-129` carries an autouse fixture `_isolate_claude_md_target` doing **both** prescribed halves — `_seed_claude_md(tmp_path)` **and** `monkeypatch.setenv("CLAUDE_PROJECT_DIR", str(tmp_path))` — reusing `_seed_claude_md` from the file this body named. It ships with a child-process proof at `:132-159` asserting the path **reported by the child**, on the rationale that only a child can witness all four links of `setenv → os.environ → subprocess inheritance → resolver`. Landed **`0be93d08`, 2026-07-23, v4.6.17** — roughly 18.5 hours after this issue was filed. It was never closed.
>
> **This issue stays open for the ROOT, which is different from what the title said.** The root is: **the Working Memory sync target is resolved ambiently, with no caller-specifiable override.** The remedy is an **explicit `target` parameter**.
>
> ⚠️ **The root is deliberately NOT phrased as "`memory_api.py:427-429` never consults `self._db_path`."** That phrasing names the defect in the vocabulary of a **rejected** remedy — coupling the sync target to `db_path`. That coupling is a production regression: one user has **one** memory DB and **many** projects, each with its own `CLAUDE.md`, so coupling them inverts an N:1 relationship. Stating the root that way would make the rejected design read as the sanctioned fix.
>
> ### Corrections to the body below
>
> **1. "Blast radius is per-checkout" is FALSE, and wrong in the dangerous direction.** Resolver branch 3 resolves `git rev-parse --git-common-dir`.parent — the **main repo root**, "whether run from the main repo or a linked worktree" (the resolver's own comment). `CLAUDE.md` is gitignored (`.gitignore:18`), so a worktree has none, branch 2 always misses, and branch 3 hits. **Running the suite inside a worktree rewrites the MAIN checkout's `CLAUDE.md`.** Worktree isolation must never be offered as a mitigation.
>
[...]

PR #1272, opened 2026-07-26
Stop the test suite writing to the live pact-memory DB, and make the sync target explicit

michael-wojcik:
## Summary

The pact-plugin test suite wrote into the developer's **live** pact-memory database on every run, and the Working Memory sync resolved its CLAUDE.md target ambiently with no caller-specifiable override. This lands the first five steps of the remediation plan.

**Two anchors, bound at different times, so no single boundary closes both:**

| | binds | via | consequence |
|---|---|---|---|
| **Anchor 1** — which DB | at **import** | `config.py`'s `Path.home()` | reach is bounded by *process identity*, not lexical scope |
| **Anchor 2** — which CLAUDE.md | at **call** | `os.environ`, through a four-branch probe-and-continue resolver | a parent-process patch cannot reach a re-importing child |

Fixing anchor 1 alone would have **hidden** anchor 2. That is why they ship together.

## Measured outcome

Leak A's live writer is closed, and the claim is **predicate-shaped and positively corroborated** rather than an absence:

- **A post-fix full-suite run adds ZERO archive-marker rows**, against a stable pre-fix **+2 per run**. That delta is the claim.
  > ⚠️ **No absolute is cited here, deliberately.** An earlier draft cited 14 as the production baseline. Decomposed and bound to a moment: **4 legitimate + 10 leaked at 05:43Z; 4 + 14 at 06:10Z** as review tooling exercised pre-fix commits. (No share is quoted — a ratio over a moving quantity drifts for the same reason the absolute does.) **An absolute over a shared, mutable production store is not a claim-bearing quantity.** If you cite one, decompose it; a re-measurer finding a different number is seeing pre-fix execution, not a regression.
- **The formerly-leaking saves now land in tmpdirs whose basenames are *exactly* the `project_id` values on the polluted rows** — same producer, demonstrably relocated. That correspondence, not a count, is what makes the closure positive rather than an absence.
  > ⚠️ **Spawn counts are NOT reconciled and are deliberately not cited as a result.** *(Final clean-tree measurement at `d35f6dec`: two probes at different install points both report **99**, blind-window **0**, closure **18 → 18** under the shipped predicate. The 99-vs-93 gap remains open and is now against a different tree again — +2 spawning tests were added in remediation.)* Two instruments measured overlapping populations and disagree: a `run`-layer wrapper reported **93 spawns / 0 unscoped**; a `Popen`-layer wrapper at HEAD reports **97 / 3 unscoped** (itself a correction — an earlier figure of 194/6 double-counted, because `run()` delegates through the replaced global). The **4-spawn and 3-vs-0 gaps are open**, and the second may be **definitional rather than factual** — the two instruments may not share a meaning for "unscoped." The three are adjudicated: two are **HOME-sandboxed** (omitting `--db-path` is the point of the test), one is **short-circuit-safe only** — `cli.py` with no args exits before any handler runs. That third is itself an inertness claim of the class residual 2 describes. **What was counted matters more than how many**, which is why both raw figures are stated and neither is asserted. Neither team re-derived the other's instrument, deliberately. **The relocation evidence above does not depend on either count.**
- Suite: **12882 passed / 15 skipped / 0 errors**, errors token scanned explicitly over whole captured output (the usual summary line drops it).

`0 new rows` alone would have been an absence, equally consistent with "isolation works" and "the save never ran." The spawn-destination evidence is what makes it a result.

## What landed

| commit | |
|---|---|
| `22bc6f08` | Bind `_cli`'s pin-context values at def time, not by closure |
| `303c3333` | Overwrite `CLAUDE_PROJECT_DIR` from an explicit cwd, not `setdefault` — **production fix**, inverted fail direction |
| `4615c603` | Require `db_path` on `build_verdict` and answer it at all 41 call sites |
| `8c9e10cd` | Refuse the production memory database on test-spawned CLI processes |
| `f272f3b9` | Make the memory-CLI target explicit at the process boundary |
| `da854d68` | Couple the explicit-`db_path` control to the check it claims to guard |
| `c89a21b1` | Cover both sync branches with the existence guard, and say so truthfully |

Steps 3 and 4 are deliberately split: step 3 is **mechanical** (41 call sites, no behavioural change) and buys *legibility*, not safety — `db_path=None` still satisfies a required parameter. Step 4 is the mechanical closure.

## ⚠️ Absolute vs delta — read before citing any row count

**`15` appears repeatedly in the commit record and is the SUBSTRING-predicate count, not the type-anchored one.**

**And the type-anchored absolute is itself a snapshot, not a constant.** It was 14 when this PR was opened and 18 shortly after — any run at a pre-fix commit increments it. This section warns against citing a stale absolute, and the first draft then shipped one. **The delta is the claim; the absolute is a timestamp.**

The spread is exactly one record that merely *mentions* the marker in an entity's name/notes rather than carrying it as an entity `type`. **Deltas were unaffected**, so no closure claim falls — but a later reader citing `15` would have no way to reconstruct why.

**Use the type-anchored predicate** for any future measurement: an entity whose `type` field *equals* `ARCHIVE_ENTITY_TYPE`, read by symbol, never hardcoded. The substring form over the serialized blob has a real false-positive class. `goal LIKE 'Archived pin:%'` is worse still — a strict subset, structurally blind to the legitimate archives.

## Open residuals — named, not discovered

> ⚠️ **This list is a predicate, not a census.** It enumerates *couplings that are true at their origin and unenforced at their point of use* — it is not a claim that there are exactly this many. A census-shaped reading is what let an untreated sibling go unlisted until review. **Pin the predicate; let the count float.**
>
> That rule has three objects in this PR alone: this finding list, the predicate spread (15 vs 14), and the production row absolute. **The delta is the claim; absolutes over shared mutable state are snapshots, not baselines.**

1. **`_project_root_of`'s two-layout knowledge is unpinned** against `claude_md_manager`'s canonical constants. Confirmed by measurement: renaming the SSOT constant broke 21 tests elsewhere while the memory layer passed 103/103 — the twins are fully decoupled and would drift silently.
2. **The 40 `db_path=None` sites encode call-site inertness *claims*** that a future edit elsewhere can falsify without touching the call. A **runtime** detector does exist (step 4's child guard converts a falsified claim into a loud failure), but for the **spawning class only** — and that coverage is *one measured site plus a call-graph inference*, not 40 measurements. What is absent is a **static** detector at review time.
3. **In-process `main()` with a real `PACTMemory` and no `--db-path`** still reaches production. Uncovered by design; the guard is scoped to spawned children. ⚠️ **This residual has NO independent confirmation.** The connect-level probe cited above (0 production connections / 60 total) is the **author's own measurement**; no reviewer re-derived it. Treat it as self-reported, not corroborated.
4. **The accusation guard in the new warning test is a bounded substring denylist** — a novel phrasing such as *"this indicates a defect"* would pass it. What carries that test is its positive half, asserting both cause-clauses are present. Bounded, not broken.

## Scope

⚠️ **The scope fence had a gap, and it is worth stating.** The plan's numbered steps were partitioned into 1–5 (in) and 6–9 (out) — but the plan also carries items under **Open Questions**, which fell into neither partition and were therefore **owned by nobody**. At least two sit in that gap, one of which is the production reachability path for a defect found during review.

Steps 6–9 are **out of scope by explicit ruling**: per-site `--no-sync` at the Leak B carrier, conftest promotion, the five-entry-point AST census, and the version bump. Their absence is deliberate, not an omission.

## Notes for reviewers

- **ARM 3** in `test_archive_pin_emit.py` is **byte-untouched across all seven commits, by design.** It is an unsuppressed negative control. If it reddens, that is a signal — not a licence to edit the control.
- **A claim in the frozen plan was falsified during implementation and deliberately not propagated.** The plan stated a total resolver in the sync path would *create* CLAUDE.md files where none existed. Testing showed it does not — the read precedes the write, so it raises and degrades. What it *does* leave is a lock sidecar. The plan's direction was right and its mechanism was wrong; the source now documents what was measured. **The consequence is the larger half:** the no-create property was an *accident of ordering*, not a contract, and `c89a21b1` converts it into a stated one.
- The census is **`ast.parse` only, never grep** — a line-oriented count returns 44 because it cannot see multi-line calls. The true figure was 41 bare of 51.
- Do not encode any count in a test. **Pin the predicate; let the count float.** A pinned count fires on every innocent addition and trains maintainers to bump the number instead of investigating.

Addresses #1265 (Leak A's live writer). For #1240: **provides the caller-specifiable override the issue prescribes. The production path still resolves ambiently, because no in-tree caller passes `target` yet** — the capability is closed, the behaviour is not. Neither issue is fully retired.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.6.20",
+      "version": "4.6.21",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -642,7 +642,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.6.20/       # Plugin version
+│               └── 4.6.21/       # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.6.20",
+  "version": "4.6.21",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.6.20
+> **Version**: 4.6.21
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/agents/pact-database-engineer.md
+++ b/pact-plugin/agents/pact-database-engineer.md
@@ -146,8 +146,8 @@ You must escalate when:
 **Self-Coordination**: If working in parallel with other database agents, check S2 protocols first. Respect assigned schema boundaries. First agent's conventions (naming, indexing patterns) become standard. Report conflicts immediately.
 
 **Algedonic Authority**: You can emit algedonic signals (HALT/ALERT) when you recognize viability threats during implementation. You do not need orchestrator permission—emit immediately. Common database triggers:
-- **HALT DATA**: DELETE without WHERE clause, DROP TABLE on production data or on any data whose restore path you have not verified, PII stored unencrypted, foreign key violations risking data integrity
+- **HALT DATA**: DELETE without WHERE clause, any DROP TABLE or other destructive write against production data or against data whose restore path you have not verified by restoring from it, PII stored unencrypted, foreign key violations risking data integrity
 - **HALT SECURITY**: SQL injection vulnerability in stored procedure, overly permissive access grants
 - **ALERT QUALITY**: Migration fails repeatedly, performance degrades significantly
 
-Read [algedonic.md](../protocols/algedonic.md) immediately on detecting a DATA-integrity threat (destructive operation without rollback, schema violation, foreign-key breach, PII exposure in unencrypted columns or logs) or any irreversible change flowing from a migration or query you are authoring, whether to production data or to any data whose restore path you have not verified.
+Read [algedonic.md](../protocols/algedonic.md) immediately on detecting a DATA-integrity threat (destructive operation without rollback, schema violation, foreign-key breach, PII exposure in unencrypted columns or logs) or any irreversible change to production data, or to data whose restore path you have not verified by restoring from it, flowing from a migration or query you are authoring.
--- a/pact-plugin/agents/pact-database-engineer.md
+++ b/pact-plugin/agents/pact-database-engineer.md
@@ -146,8 +146,8 @@ You must escalate when:
 **Self-Coordination**: If working in parallel with other database agents, check S2 protocols first. Respect assigned schema boundaries. First agent's conventions (naming, indexing patterns) become standard. Report conflicts immediately.
 
 **Algedonic Authority**: You can emit algedonic signals (HALT/ALERT) when you recognize viability threats during implementation. You do not need orchestrator permission—emit immediately. Common database triggers:
-- **HALT DATA**: DELETE without WHERE clause, DROP TABLE on production data, PII stored unencrypted, foreign key violations risking data integrity
+- **HALT DATA**: DELETE without WHERE clause, DROP TABLE on production data or on any data whose restore path you have not verified, PII stored unencrypted, foreign key violations risking data integrity
 - **HALT SECURITY**: SQL injection vulnerability in stored procedure, overly permissive access grants
 - **ALERT QUALITY**: Migration fails repeatedly, performance degrades significantly
 
-Read [algedonic.md](../protocols/algedonic.md) immediately on detecting a DATA-integrity threat (destructive operation without rollback, schema violation, foreign-key breach, PII exposure in unencrypted columns or logs) or any irreversible change to production data flowing from a migration or query you are authoring.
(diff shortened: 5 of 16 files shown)


michael-wojcik:
## Measurements that exist nowhere in the commit record

These were taken during verification and live only in session scratch. Recording them here so reviewers can ask precisely, and so the "before" half of each delta survives.

### Pre-fix baseline at `da854d68` — the "before" of the F1 delta

```
ARM 1  explicit+absent   -> False   leftovers []                    DEBUG    "explicit sync target ... does not exist"
ARM 2  ambient+total     -> False   leftovers ['.CLAUDE.md.lock']   WARNING  "Failed to sync ... [Errno 2] No such file"
ARM 3  LIVE CONTROL      -> True    wrote marker                    INFO     "Synced memory to CLAUDE.md ..."
```

Post-fix, ARM 2 shows `leftovers []` and the specific new message. **The WARNING level alone does not discriminate** — the generic degradation handler already emitted one pre-fix. The sidecar and the message text are what carry the claim.

### Mutation cardinalities

| mutation | result |
|---|---|
| child-guard early-return removed | unscoped spawn proceeds, no refusal |
| a real short-circuit removed above a `db_path=None` site | loud refusal (`UNSCOPED_TEST_DB`), no store opened |
| explicit-target guard removed | sidecar returns |
| SSOT layout constant renamed | **21 fail elsewhere / 103 pass in the memory layer** |
| `setdefault` restored | 2 fail |
| `env=None` restored | 1 fail |
| guard moved back inside the explicit branch | 2 fail / 33 pass |

The layout-rename result is the evidence for residual 1: the twins are **fully decoupled**, so a rename drifts silently.

### Runtime probes

- **0** production sqlite connections against **60** total, with positive-control basenames confirming the wrapper reached the memory layer. Wrapping `sqlite3.connect` rather than `PACTMemory.__init__` was deliberate — wrapping the class enumerates a *construct*, which is the error that produced a separate detector's false zero.
- **194** memory-CLI spawns, `phase_counts {'call': 194}`, **0** in the blind window. The guard's documented gap (`PYTEST_CURRENT_TEST` is popped between items) is real by design and **unoccupied in practice**.
- `_resolve_display_claude_md_with_base()` returns `(None, None)` from an isolated non-git tmp dir with no CLAUDE.md and no ambient var. This is the **premise** under the DEBUG-vs-WARNING split, and it had been taken from the resolver's docstring by everyone until it was run.
- Predicate spread: **substring 15 vs type-anchored 14** on the same store at the same instant.

### Evidence scoping — stated so it is not overread

- **41 → 0 bare call sites** is an **author-and-auditor pair**, not three independent measurements. The auditor's census was banked at base `0ce19db8` *before any implementation existed* and re-run against a hermetic `git archive` export, which is what makes the pair meaningful — but it is a pair.
- **Residual 2's runtime coverage is one measured site plus a call-graph inference**, not 40 measurements. The inference: `build_verdict`'s only route to a store is `_run_memory_cli`, which spawns, and the child guard covers spawns.
- Of the four structural couplings, **only the sync guard was ever a live overclaim**. The other three are latent.

Issue #1329, opened 2026-07-30
test_deny_rows_minted_first fails intermittently in the full suite (~1 in 3), passes in isolation


michael-wojcik:
## Summary

`tests/test_merge_guard_obs_cert.py::TestObsHRealMintExecuteRetireRoundTrip::test_deny_rows_minted_first` fails **intermittently** when the full suite runs — observed in 1 of 3 consecutive full runs — and passes reliably when run in isolation.

Surfaced by the concurrent auditor during at-rest verification of the `feat/1175-pin-cap-boundedness` branch.

## Not caused by that branch

Established three independent ways:

1. **Collection order** — the test collects BEFORE both of that branch's new test files, so those files cannot influence it by ordering.
2. **Isolation** — passes alone, and passes in its own module (272 passed, twice).
3. **No module intersection** — zero overlap between the 13 files the branch changes and the four merge-guard modules the test imports.

## A confounded inference, recorded because it nearly shipped

An **exclusion run** — full suite minus the two new test files — came back fully green, which reads as "the new tests are implicated". A subsequent full run **with them present** passed too, falsifying that reading.

The exclusion result was **confounded by the flakiness itself**: with a ~1-in-3 failure rate, a single green run is not evidence of causation. Any future attempt to bisect this by excluding files needs repeated runs per arm, not one.

## What is not yet known

- The shared state responsible. `deny_rows_minted_first` suggests ordering or a mint/retire round-trip touching state another test also writes, but this was **not** traced.
- Whether the failure rate depends on `-p xdist` / worker count, or on which other modules happen to run first.

## Suggested next step

Re-run the full suite N times capturing seeds and worker assignment, and diff the passing and failing environments for that test. Do **not** conclude a cause from a single green arm.

Issue #1328, opened 2026-07-30
Teachback review: mechanically sum a proposed per-dimension variety re-score against its stated total and band


michael-wojcik:
## Summary

When a teammate's teachback proposes a per-dimension variety re-score, nothing checks that the four proposed digits actually sum to the total the teammate claims. A wrong sum can silently move a dispatch across a routing band — including a band that changes the gate applied to the very message making the claim.

Add a mechanical check at teachback review: **sum the four proposed dimension values, compare against (a) the stated total and (b) the band the dispatch is currently in.**

## Observed instance

A teachback proposed raising one dimension and wrote:

> "Novelty 3 plus Scope 3 plus Uncertainty 3 plus Risk 2 also totals 10, so the routing is unchanged and nothing depends on the correction."

`3 + 3 + 3 + 2 = 11`, not 10.

The diagnosis in that teachback was **correct** — the rationale it flagged was genuinely defective. Only the arithmetic was wrong. That is what makes this worth mechanising: the surrounding reasoning was sound enough to be persuasive, and the sentence asking to be trusted was the one carrying the error.

## Why the consequence is not cosmetic

From `pact-plugin/hooks/shared/variety_scorer.py`:

- `ORCHESTRATE_MAX = 10`
- `PLAN_MODE_MIN = ORCHESTRATE_MAX + 1` (= 11)
- `route_workflow` returns `orchestrate` at 10 and `plan-mode` at 11

So the proposed re-score does not leave routing unchanged; it moves the dispatch into the next band.

**Second-order consequence, which is the sharper one:** the L1.5 gate makes `reasoning_reconstruction` REQUIRED at variety 11+ and optional below. At 11 the proposed re-score would have retroactively made that field mandatory on the very message proposing it.

**A re-score can change the gate applied to itself.** That property is not obvious from either the scorer or the gate read alone.

## Why this one is safe to mechanise

- It needs no judgement: four integers and two comparisons.
- It has no false-positive mode that costs a user anything. It cannot block a faithful action; it annotates a claim.
- It fires exactly on the sentence that is asking to be trusted without a check.

## Suggested shape

At teachback review, when `metadata.teachback_submit.variety_acknowledgment` proposes dimension values:

1. Sum the four proposed values.
2. If the sum differs from any stated total in the same payload, surface the discrepancy.
3. Report the band the proposed sum lands in, and whether it differs from the current band.
[...]

Issue #1327, opened 2026-07-30
migrate_to_managed_structure bounds an unbounded pinned region without excluding the absorbed entries


michael-wojcik:
## Summary

`_build_migrated_content()` (via `migrate_to_managed_structure()`) converts an unmigrated `CLAUDE.md` whose `## Pinned Context` region has **no terminating H2** into a managed file that is **bounded but wrong**: the entries the missing terminator had absorbed into the region stay inside `## Pinned Context` and continue to count as pins.

**"Bounded by construction" is not the same claim as "correctly bounded."**

## Measured

Fixture: a `CLAUDE.md` with 2 real pins and 12 dated Working Memory `### ` entries, with no H2 terminating the pinned region.

| input | after `_build_migrated_content` |
|---|---|
| unmigrated, unbounded | `bounded=True`, **pins=14** |

The 12 Working Memory entries are absorbed as pins. Cap is 12, so the file is over cap on the count axis by construction.

## Why this is latent today

Such a file denies before migration and denies after it, so the migration changes nothing observable. There is no user-visible symptom at present.

## Why it becomes decisive

Issue #1175 PR 1 installs a boundedness gate that **declines enforcement** when the pinned region is unbounded (fail-open). That fix is removed by this defect:

1. PR 1's gate sees `bounded=False` and declines. Curator is correctly allowed to edit.
2. `migrate_to_managed_structure()` runs on the next SessionStart.
3. The file becomes `bounded=True` with the 12 phantoms still inside.
4. The gate no longer declines, because the file is no longer unbounded. The deny returns.

The migration is the mechanism #1175 nominates as the thing that drives the unbounded population toward zero. As written it would silently undo the fix it is supposed to support.

## Scope

One production call site: `pact-plugin/hooks/shared/claude_md_manager.py:1016`, inside `migrate_to_managed_structure()`. Tree-wide there are 101 references to `_build_migrated_content`; every other one is a test.

#1175 PR 1 routes around this by ordering `ensure_pinned_terminator()` **before** `migrate_to_managed_structure()` at `session_init.py:900` (measured: migration alone gives 14 pins, repair-then-migrate gives 2). That ordering is a PR 1 constraint, not a fix for this issue.

**This issue is about the function's own correctness**, which the ordering does not address.

## Suggested direction

[...]

PR #1330, opened 2026-07-31
Stop the pin caps enforcing against a phantom count

michael-wojcik:
Closes part of #1175. This is **PR 1 of the pin-cap redesign arc** — the boundedness infrastructure and the over-block fix. It does not change any cap value.

## The defect

The pinned region of a project `CLAUDE.md` is terminated by a `## Working Memory` heading. When that heading is missing, the parser scans past the region and counts ordinary `### ` entries as pins. The caps then enforce against a **phantom count**: a curator holding 2 real pins is denied at 15/12 and told to run `/PACT:prune-memory` to demote pins they do not have.

That is an **impossible cure** — the deny names a remedy the reader cannot perform.

Three distinct deny paths reach it, not one:

| Arm | Deny reason |
|---|---|
| Ordinary note added | `embedded_pin` |
| Legitimate pin added | `count`, 15/12 |
| Long prose added | `size` |

The embedded-pin check returns before either cap axis is consulted, so a guard placed around the cap logic alone leaves that arm live.

## What this PR does

1. **`_parse_pinned_section` reports boundedness.** Returns a `PinnedSection` NamedTuple; 16 call sites migrated.
2. **The gate declines enforcement when the region is unbounded**, before the call to `compute_deny_reason` so all three arms are covered. The decline records which side was unbounded, so it is observable. A bounded region over the cap still denies.
3. **The missing terminator is repaired at SessionStart**, placed before the first undated heading — a terminator at the *end* of the region reproduces the defect it exists to prevent. The repair is not frame-scoped, because the managed-structure migration it must precede has never been frame-scoped; scoping only the repair made the outcome depend on which frame type opened a session first, with the losing branch permanent.
4. **The embedded-pin deny message names both remedies**, and does not promise the edit will then be allowed.
5. **The fresh-start write arm** is gated on the written region's boundedness alone. A genuinely above-cap write still denies.
6. **The repair refuses** when the managed region already contains the terminator text, and says what it actually found rather than what it assumed.

## Certification

Every behavioural claim is certified **bidirectionally** — green on correct input, red on the defect — never by an additive-diff argument.

- **Eighteen pre-registered revert probes**, each registered before it ran. One prediction failed and is recorded rather than dropped.
- A concurrent auditor verified every commit **at rest** off `git archive` exports rather than the live tree, and re-derived an AST source-segment hash to confirm the out-of-scope `_parse_working_memory_section` is byte-unchanged.
- Suite on **Python 3.14.6**: 13,239 passed, 15 skipped, 0 failed, 0 errors, with an explicit errors-token scan. Re-run on **Python 3.13**, which is what CI gates on: **13,234 passed, 19 skipped, 0 errors**, plus one load-sensitive timing test in the merge-guard suite that passes 3/3 in isolation and shares no file with this change. Naming the interpreter because the first two attempts to run on 3.13 reported exit code 0 while executing zero tests.

Checks were found that **could not fail**, and were repaired or replaced. The count is boundary-dependent — an enumeration afterwards ran from 8 to 17 across four defensible readings — so the boundary is stated rather than left implied.

**Eight**, counted across this arc's certification artifacts: the shipped test suite AND the design document and its measurement harnesses. Some of those artifacts are gitignored and ship with nothing; they are counted because the checks and the repairs were real, not because they ship. Excluded: one row that named a discriminating verdict but paired it with a non-discriminating count assertion, so a coder could satisfy it with the weaker half — its control can fail, so it is not a member of this set. The class is checks that could not fail wherever they lived, so a defect in the design document repaired by a mechanism in the shipped suite still counts — the repair locus need not match the defect locus. Also excluded: any check an agent caught and replaced inside its own turn before another party saw it. Two examples of the eight:

- The full-file boundedness formula `end < len(content)` is **true on 100% of marked files**, bounded or not — a wrong implementation would have shipped looking complete while doing nothing. Both natural fixtures force the agreeing branch, so every boundedness fixture now asserts the region was found *before* asserting it is bounded.
- The decline test ran a five-pin fixture against a cap of twelve, so deleting the decline left its verdict assertion passing; the classification assertion beside it carried the whole test. The fixture is now sized from the cap so both halves discriminate.

The last of these found a live defect: covering the second reachable input class of the nothing-absorbed guard catches a repair that inserts the terminator *ahead* of a genuine pin and **under-counts** — the mirror image of the over-count this PR removes.

## Scope

Out of scope by ruling: `working_memory.py`'s first-match section lookup (belongs to pact-memory), and `commands/pin-memory.md`, deferred to PR 2.

Known follow-ups: #1327 (the migration bounds an unbounded region without excluding absorbed entries), #1328, #1329.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "PACT",
       "source": "./pact-plugin",
       "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
-      "version": "4.6.24",
+      "version": "4.6.25",
       "author": {
         "name": "Synaptic-Labs-AI"
       },
--- a/README.md
+++ b/README.md
@@ -642,7 +642,7 @@ When installed as a plugin, PACT lives in your plugin cache:
 │   └── cache/
 │       └── pact-plugin/
 │           └── PACT/
-│               └── 4.6.24/     # Plugin version
+│               └── 4.6.25/     # Plugin version
 │                   ├── agents/
 │                   ├── commands/
 │                   ├── skills/
--- a/pact-plugin/.claude-plugin/plugin.json
+++ b/pact-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "PACT",
-  "version": "4.6.24",
+  "version": "4.6.25",
   "description": "Orchestration harness that turns Claude Code into a coordinated team of specialist AI agents",
   "author": {
     "name": "Synaptic-Labs-AI",
--- a/pact-plugin/README.md
+++ b/pact-plugin/README.md
@@ -1,6 +1,6 @@
 # PACT — Orchestration Harness for Claude Code
 
-> **Version**: 4.6.24
+> **Version**: 4.6.25
 
 Turn a single Claude Code session into a managed team of specialist AI agents that prepare, design, build, and test your code systematically.
 
--- a/pact-plugin/hooks/pin_caps.py
+++ b/pact-plugin/hooks/pin_caps.py
@@ -407,10 +407,26 @@ DENY_REASON_SIZE = (
     "if the content is verbatim load-bearing."
 )
 
+# AUGMENTED, not replaced. The old text named ONE remedy — "use `#### ` or
+# bold" — which is right only when the heading was meant as in-body structure.
+# When the curator meant to add a PIN, that advice prescribes the very
+# construct that makes the block un-prunable (a `#### ` block never becomes a
+# Pin, so /PACT:prune-memory cannot reach it), and it gives no path at all
+# from the deny to the fix. Both remedies are named below, each with the
+# question that selects it.
+#
+# IT DELIBERATELY DOES NOT PROMISE THE EDIT WILL THEN BE ALLOWED. Adding the
+# date comment turns the heading into a real pin, which consumes a slot — so
+# a curator at the cap converts an embedded-pin deny into a count deny. The
+# text says what to DO, never what will HAPPEN.
 DENY_REASON_EMBEDDED_PIN = (
-    "Candidate body contains an embedded pin structure "
-    "(a `### ` heading). On reload this would be counted as an extra pin "
-    "and defeat the count cap. Use `#### ` or bold for in-body structure."
+    "Candidate body contains an embedded pin structure — a `### ` heading "
+    "with no `<!-- pinned: DATE -->` comment on the line above it. On reload "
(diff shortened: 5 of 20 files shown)


michael-wojcik:
## Closing as superseded — the work is retained, not discarded

Superseded by the user's rulings of **2026-08-01** on the five open decisions in `docs/plans/pin-cap-replan-plan.md`. The ruling set is recorded in the body of #1175.

**Nothing here is lost.** Closing a pull request does not delete its branch. The 14 commits are retrievable by three independent paths, each verified:

| Path | Location |
|---|---|
| fork branch | `michael-wojcik/PACT-Plugin` `refs/heads/feat/1175-pin-cap-boundedness` |
| **PR ref (permanent)** | `Synaptic-Labs-AI/PACT-Plugin` `refs/pull/1330/head` — GitHub retains this even if every branch is deleted |
| head SHA | `8d8b5aea92327ede641b647da286e2417d8f3d78` |

> **CORRECTED IN PLACE.** An earlier version of this comment gave the branch name without naming its remote. **The branch is NOT on `origin`** (the upstream `Synaptic-Labs-AI` repo) — it is on the `michael-wojcik` fork, which is where head branches are pushed in this project. A reader running `git ls-remote --heads origin` would have found nothing and concluded the work was gone. Use `refs/pull/1330/head` on upstream, which needs no fork access and cannot be deleted.

**What this branch holds, and why the rulings retire it.**

Its central artefact is the pinned-terminator repair writer (`_REPAIR_REFUSED_NO_DATED_PIN`, `_REPAIR_REFUSED_NOTHING_ABSORBED`, `_REPAIR_REFUSED_PARSER_DISAGREEMENT` in `claude_md_manager.py`, which is 1467 lines on this branch against 1295 on `main`), plus `test_pinned_terminator_repair.py`. That writer exists only here — it was never on `main`.

It was authorised by `RULING 5` of 2026-07-30, and that ruling is now **retired and replaced** on three independent grounds:

1. Its code is not on `main`, so re-affirming it would commission a fresh build rather than preserve running work.
2. Its condition has **zero instances** — two lanes measured 137 files independently and found no unbounded pinned region.
3. Its first constraint, *"terminator only, never pin content, headings or ordering"*, is unsatisfiable as worded: on a file where the pin set cannot be identified, where the terminator goes decides which entries are pins.

The branch's `bounded` inference is retired separately by ruling 1, which replaces an inferred region boundary with declared marker pairs — there is no terminator left to guess.

**This branch has not been audited for salvageable pieces.** The two artefacts named above are ruled out. Anything else on it is retrievable at the SHA above if a later phase wants it.

**Do not treat this branch as a foundation for the replacement work** (user instruction, 2026-08-01).

Issue #1175 stays OPEN.

Issue #1370, opened 2026-08-02
'pytest tests/' silently drops 48 tests that live inside the skill directories they guard


michael-wojcik:
## Summary

Two test files live **outside** `pact-plugin/tests/`, **inside the skill directories they test**:

- `skills/pact-agent-teams/test_skill_loading_agent_teams.py` — 26 tests
- `skills/pact-handoff-harvest/test_skill_loading_harvest.py` — 22 tests

A `pytest tests/` invocation **never collects them**. Measured on one commit, one tree:

```
pytest --collect-only          -> 13595
pytest tests/ --collect-only   -> 13547
```

**48 tests, and they are the ones sitting next to the files most likely to be edited.**

## Why this is a gate-integrity problem, not a trivia item

PACT's own merge-gate guidance says *"full suite as the merge gate — plain `pytest`"*. But `pytest tests/` **looks** like the full suite. It is the obvious invocation, it prints a five-figure passed count, and it produces a green that **never looked at the tests adjacent to the thing being changed.**

The failure mode is precisely inverted from what an operator would expect: **the more directly a change targets a skill, the more likely its guarding tests are the ones the scoped invocation drops.**

This is the same shape as the pin-coverage gap in the sibling issue: **a guard whose scope does not cover its subject.** There, the pin's key did not match the convention's key. Here, the invocation's scope does not include the change's neighbourhood.

## How it surfaced

Two parties measured collection counts on the same commit and disagreed. One had passed a `tests/` path argument, the other had not.

The discrepancy was initially attributed to an **export effect** — the hermetic `git archive` used for one set of runs strips `.git`, which genuinely does cause a large skip delta elsewhere in the same session. That was a **plausible, well-supported, and wrong** mechanism: the export collects all 13,595 without a path argument.

The correct cause was found only by insisting the two numbers reconcile rather than attributing the gap to a known hazard. Worth recording because the wrong explanation was the more sophisticated one.

**Nothing was missed on the change that surfaced this** — both files were checked and contain no assertions relevant to it. The hazard is general.

## Remedy options

**A — make the scoped and unscoped invocations agree.** Move the two files under `tests/`, mirroring the directory they cover. Then no invocation silently defines a smaller suite. Simplest, and it removes the trap rather than documenting it.

**B — state the gate invocation explicitly** wherever the full-suite rule appears, and say that a path argument narrows it. Cheap, and it relies on every future operator reading and remembering.

[...]

Issue #1368, opened 2026-08-02
Auditor RED routing: orchestrate.md tells the auditor to pause a coder, which its agent body and pact-audit.md both forbid


michael-wojcik:
## Summary

Three shipped surfaces describe what a concurrent auditor does on a RED signal. **Two agree. One does not, and it is the one the orchestrator reads aloud into every auditor dispatch.**

| surface | says |
|---|---|
| `agents/pact-auditor.md:59` | "**Do NOT direct coders** — Ask questions, do not give instructions. **Report to orchestrator, not to coders**" |
| `agents/pact-auditor.md:102` | "Never direct coders \| Report to orchestrator. **The orchestrator decides whether to intervene**" |
| `protocols/pact-audit.md:31–32` | `SendMessage to coder` → "Only when file observation raises a question code alone can't answer" · `RED signal to orchestrator` → "Clear architecture violation or requirement misunderstanding" |
| **`commands/orchestrate.md:767`** | "**On RED signal: SendMessage to the affected coder and pause their work.**" |

The agent body and the protocol are coherent with each other: **coder messages are for questions; RED goes to the orchestrator.** The dispatch template is the outlier.

## Why this is not cosmetic

The outlier does not merely re-route a message — **it grants the auditor an authority the other two surfaces explicitly reserve for the orchestrator.** "Pause their work" is an intervention. `pact-auditor.md:102` states the opposite in the same breath as the rule: *the orchestrator decides whether to intervene.*

So the two readings diverge **exactly at the moment they matter most** — a real RED, mid-implementation, under time pressure — and they diverge on **who may stop a running teammate**.

There is also a coordination hazard the agent body's version avoids. A lead may have a ruling in flight that the auditor cannot see. An auditor that pauses a coder directly can collide with it, and neither party learns that it happened.

## How it surfaced

A concurrent auditor received a dispatch built from the `orchestrate.md` template, noticed the conflict with its own agent body, **and asked rather than silently choosing**. It applied the agent-body rule and stated the default it was applying.

**The resolution is untested.** No RED arose during that phase, so the direct-to-coder path was never exercised. The contradiction is still present in both surfaces, and the next auditor has to re-derive the same answer — or, under pressure, not notice at all.

The auditor also cited two issue numbers as prior confirmations of the rule. Neither resolves to this subject. **The phenomenon is real and those references do not support it** — which is itself a reason to fix the text rather than rely on operators remembering a convention.

## Remedy

[...]

Issue #1364, opened 2026-08-02
pact-orchestrator.md states a 25KB cap for auto-memory: wrong unit, and it is unmeasured whether any byte cap applies


michael-wojcik:
## Summary

`agents/pact-orchestrator.md` tells the orchestrator that **auto-memory** — the platform's project-scoped `~/.claude/projects/{hash}/memory/MEMORY.md` — is capped at *"the first 200 lines / 25KB"*.

Two things are wrong with that sentence, and **the second is the reason this is an open question rather than a patch**.

## Defect 1 — the unit is wrong, and it is this repository's own root-cause defect

The enforced cap in Claude Code bundle 2.1.220 is `PRe = 25000` **UTF-16 code units**, measured on trimmed text via `byteCount: t.length` — a field whose *name* is a misnomer, because JS `String.length` counts UTF-16 code units, not bytes.

`25000 / 1024 = 24.41`, which is why the platform's truncation warning renders as `24.4KB`. **The unit is destroyed at render time**, and every downstream reader who re-parses that rendered string recovers either 24,400 or 24,986 — neither of which is 25,000.

Writing the cap as **"25KB"** is that same class of error, one step earlier: a code-unit count stated in a byte unit. It is the defect the agent-memory-headroom work exists to correct, sitting in PACT's own instruction text.

## Defect 2 — the cap may not apply to this system at all, and that is unmeasured

**This is the blocking uncertainty.** Do not fix the unit until it is resolved.

The agent-memory headroom work measured the cap for **`~/.claude/agent-memory/`**. It never measured **auto-memory**, which is a different system at a different path.

Two observations pull in opposite directions:

- **Toward "same cap"** — the enforcing code path is keyed on the filename `PS = "MEMORY.md"` with the shared `fie` (200 lines) and `PRe` (25,000 units) constants. Auto-memory's index is also named `MEMORY.md`.
- **Toward "no cap"** — the delivered **project-memory** system-prompt block states **no cap at all**, while the delivered **agent-memory** block states the 200-line one. Two different agent types were sampled and both agree on that asymmetry.

So the honest position is: **the enforcement relationship between the shared code path and auto-memory's `MEMORY.md` has not been observed.** Correcting `25KB` to `25,000 UTF-16 code units` would assert that relationship rather than establish it.

[...]

Issue #1295, opened 2026-07-27
CI collects 12,928 of 12,976 tests — the 48-test gap holds the only pins on two shipped SKILL.md files


michael-wojcik:
## Summary

**CI runs `python -m pytest tests/`, which collects 12,928 of the repo's 12,976 tests. The 48-test gap contains the only pins on two shipped `SKILL.md` files.** A PR that edits one of those skills passes CI without ever running its guards.

Measured on `fix/1286-wrap-up-teammate-shutdown` @ `04ca6553`:

```
cd pact-plugin
python -m pytest --collect-only -q        ->  12976 tests collected
python -m pytest tests --collect-only -q  ->  12928 tests collected
                                              ------------------------
                                              gap:     48
```

The gap, enumerated:

| file | tests |
|---|---|
| `skills/pact-agent-teams/test_skill_loading_agent_teams.py` | 26 |
| `skills/pact-handoff-harvest/test_skill_loading_harvest.py` | 22 |
| | **48** |

## The invocation

`.github/workflows/tests.yml:131`

```yaml
run: python -m pytest tests/
```

This is deliberate, not accidental — the file states the intent twice:

- `:27` — "the canonical `cd pact-plugin && python -m pytest tests/` invocation"
- `:128` — "Equivalent to `cd pact-plugin && python -m pytest tests/`"

So the fix is not "someone typo'd a path". The *canonical invocation itself* is narrower than the test corpus, and the workflow faithfully implements it.

## Why this is worse than a coverage gap

**`tests/` looks like the corpus and is not.** Every signal points the wrong way: the directory is named `tests`, the workflow comment calls the invocation canonical, the run is green, and the count is five digits. Nothing in the output distinguishes "12,928 passed" from "12,976 passed". A contributor has no reachable evidence that 48 tests exist outside the boundary.

The tests in the gap are not incidental. They are the **only** pins on their `SKILL.md` files — the guards that would catch a skill surface being edited into an inconsistent state.

## Live instance, from the change that surfaced this

The branch above edits `skills/pact-agent-teams/SKILL.md`. Its pins live in `skills/pact-agent-teams/test_skill_loading_agent_teams.py` — **inside the gap**:

```
:106  def test_has_shutdown_section(self, skill_content):
:162  def test_references_shutdown_request(self, skill_content):
:163      """Skill must reference shutdown_request for graceful termination."""
:164      assert "shutdown_request" in skill_content, (
```

[...]

PR #1371, opened 2026-08-02
Report what the memory layer actually did, instead of a signal true about something narrower

### LINKED ISSUE #1047: fix(ci): pact-memory CLI subprocess tests flake — every save/update cold-installs deps via pip, racing the 60s timeout
## Summary

`TestCliSubprocess` / `TestCliSaveVerificationE2E` in `pact-plugin/tests/test_memory_cli.py` flake **red on `main`** on cold CI runners. Run [28275162571](https://github.com/Synaptic-Labs-AI/PACT-Plugin/actions/runs/28275162571) on `main` failed with **12 `subprocess.TimeoutExpired` after 60s** on `cli.py save`/`update` (zero assertion failures); an immediate re-run passed clean (**9688 passed, ~118s**). The identical content had passed on the PR branch minutes earlier — the tell is the duration: **816s (failed) vs ~118s (re-run)** ≈ 12 × 60s of stacked timeouts.

## Root cause (confirmed)

Every `save`/`update` → `PACTMemory.save()/.update()` → `_ensure_ready()` → `ensure_memory_ready()` → `check_and_install_dependencies()` (`pact-plugin/skills/pact-memory/scripts/memory_init.py:41-86`), which runs `subprocess.run([python, '-m', 'pip', 'install', '-q', pkg], timeout=60)` for each missing dep (`pysqlite3`, `sqlite-vec`, `model2vec`).

CI installs none of these — `.github/workflows/tests.yml:94` installs only `pip pytest hypothesis httpx pytest-asyncio` — so **each fresh CLI subprocess cold-installs the full `model2vec` dependency tree** (numpy, huggingface_hub, tokenizers, safetensors, …). Under network variance the cumulative install exceeds the test's outer `timeout=60` (`test_memory_cli.py:~1402+`) → `TimeoutExpired`. `_initialized` is per-process, so every subprocess re-pays the cold cost (it self-heals once the shared pip cache is warm — hence the green re-run).

The model2vec HuggingFace download (`embeddings.py:54-55`) is a **secondary** cost, gated behind `SQLITE_EXTENSIONS_ENABLED` (`memory_api.py:459-465`), which is False in CI when deps are absent — so it isn't even reached. There is **no** offline/cache handling (`HF_HUB_OFFLINE` / `HF_HOME` / vendored model path are absent from code and workflows).

The in-process equivalents (`TestCliSaveCommand`) pass because they mock `scripts.cli.PACTMemory` — only the black-box subprocess tests hit the real auto-install path.

## Relationship to #98 — SAME ROOT CAUSE

**#98 ("change pact-memory auto-install to warn-only") targets the exact same function** (`check_and_install_dependencies`). #98 was filed for a consent / invasive-pip reason, but implementing it (warn-only auto-install, or gating install behind an opt-in env var) **eliminates this flake as a side effect**: no auto-install → `save`/`update` return immediately, embedding silently skipped via the existing `SQLITE_EXTENSIONS_ENABLED` early-return. **This issue tracks the CI-reliability symptom; the fix is #98.** It also raises #98's priority — the auto-install is now actively reddening `main`, not just a UX concern.

## Affected tests (12)

All in `tests/test_memory_cli.py` (`TestCliSubprocess`, `TestCliSaveVerificationE2E`); the hung command is always `save` or `update` (list/search/delete appear because their setup `save` hung):

- `test_update_unique_prefix_resolves_and_mutates`
- `test_update_full_hash_unchanged`
- `test_update_success_envelope_returns_resolved_full_id`
- `test_uppercase_full_id_resolves_to_lowercase_stored_update`
- `test_update_and_verify`
- `test_save_via_stdin`
- `test_save_roundtrip_confirms_verification_passed`
- `test_save_via_stdin_confirms_verification_passed`
- `test_list_returns_saved_memory`
- `test_list_limit_flag`
- `test_search_returns_results`
- `test_delete_and_verify`

## Fix options

1. **Warn-only auto-install (implements #98) — RECOMMENDED.** `memory_init.py:41-86` → report missing deps instead of pip-installing (or gate install behind opt-in `PACT_AUTO_INSTALL=1`). CI subprocesses then never pip-install → `save`/`update` return fast and deterministically. Also update the auto-install behavior tests in `tests/test_memory_init.py` (they assert install/timeout semantics and must move to the warn-only contract). Minimal root-cause fix; closes #98.
2. **Pre-install + cache deps and the model in CI.** `tests.yml`: add `pysqlite3-binary sqlite-vec model2vec` + `actions/cache` for `~/.cache/pip` and `~/.cache/huggingface` + a model warm-up step. Heavier; adds a network dependency (the workflow header already flags the native-dep path as a known flake source) and doesn't fix the invasive auto-install for real consumers.
3. **Test-scoped stopgap.** Env-skip the auto-install in the subprocess tests (requires the Option-1 env guard), or raise/remove the 60s timeout, or `skipif` when deps aren't importable. Interim mute only — fragile.

## Acceptance

- The subprocess CLI tests are deterministic on a cold CI runner (no `pip install` in the hot path).
- `main` no longer flakes red on a cold-cache run.

_Discovered while verifying CI after merging the live-probe ledger PR #1046 (which itself is unrelated — its content passed identically on the PR branch)._

### LINKED ISSUE #1346: Secretary memory-index headroom is on a measurable slope and crosses its escalation threshold next session
## Summary

The secretary's memory index is approaching a headroom threshold on a measurable trend, and the signal currently exists only in a session message channel — which has no durability tier by construction.

**Measured this session (2026-08-01, team `session-87bc0ef1`):**

- **Nine compaction requests in a single session.**
- Headroom **1,970 bytes** against a **~1,500-byte escalation threshold**.
- Falling roughly **180 bytes per session**, with **entry count flat** — so entries are growing, not multiplying.
- On that trend the threshold is crossed **next session**.

## Why this is filed rather than left to the next session

The secretary flagged it deliberately, described it as planned rather than sudden, and stated it would stop re-running the arithmetic and escalate a structural fix once the threshold is crossed. That is the right disposition and this issue does not pre-empt it.

What this issue captures is the **trend data**, which would otherwise live only in a `SendMessage`. The same session established that the message channel has no durability tier by construction: a `SendMessage` emits no journal event, is not mirrored by `task_metadata_snapshot`, and is not an artifact. A next-session secretary would arrive at the threshold without the slope, the entry-count-flat observation, or the nine-request count that shows the rate.

## Diagnosis and remedy belong to the secretary

Deliberately not proposed here. The observation that matters for whoever picks it up is that **entry count is flat while size grows**, which points at per-entry growth rather than at accumulation — so a remedy that prunes entries may not move the number, and the fix is more likely structural.

## Related

Same session filed #1344 (dispatch enumerations reading as censuses) and #1345 (the hookable-tool-set claim and its wrong-population remedy). This one shares only the durability observation, not the failure mode.

michael-wojcik:
Closes #1346 (items 1–3; item 4 deferred). Also closes #1047, #1280, #1295, #1364, #1368, #1370, #1372, #1377, #1378, #1382, #1241, #1244, #1251.

Two arcs. The branch began as an agent-memory instruction reconciliation. It then absorbed a defect family found while verifying it: **a signal true about a narrower proposition than the one being read for.**

## Arc 1 — agent-memory instructions

Reconciles PACT's `~/.claude/agent-memory/` instructions with what the platform actually delivers. `~/.claude/pact-memory/` is untouched.

The issue as filed measured the right file against a limit that never existed, in a unit that was never bytes. `"24.4KB"` is `25000 / 1024` — **25,000 UTF-16 code units rendered in KiB**. The unit is destroyed at render time, so no back-parse of the warning string recovers it. Corrected: zero of 39 indexes truncate, and the escalation threshold is not crossed.

The size-ceiling row **inverted during review** — the platform states only the 200-line limit, so the row became ADD rather than DELETE.

## Arc 2 — the narrow-signal family

**A save reported success over a missing vector.** The verification read back the memory row only, so it could not detect a missing vector by construction. The store path now reports its own capability, through the API and the CLI envelope, for save and update.

**A refused sync and a suppressed sync were the same observation.** A bare boolean meant both presented as "the file is unchanged" — which silently emptied a negative control that kept passing while measuring nothing. Now a result carrying `wrote / refused / suppressed / unresolved / missing / failed`, with `__bool__` preserving the previous meaning exactly so no caller changes behaviour. Deliberately **not** the sibling convention, which is truthiness-inverted; copying it would have flipped every truthiness read.

**A refusal wrote one free-text line into a structured channel.** The refused-sync handler logged at WARNING. The package configures no logging handler, so `logging.lastResort` carries WARNING and above to stderr — and stderr is where this CLI writes the error envelope callers parse as JSON. The fact was never being swallowed: the handler sets `sync_status='refused'` one line above and the CLI surfaces it on stdout, so the log line was redundant for machine consumers and corrupting for them at the same time. Dropped to DEBUG, with the reason stated at the site rather than 111 lines below where a maintainer could not see it. Found because a certification was read past its own commit: the fix was certified at one SHA, and a later commit on the same branch reopened the class at a different file.

**The containment check could not validate its own anchor.** Both callers derived the anchor from the same resolution that produced the target, so it asked whether a directory contains itself. A caller may now declare it: absent means today's behaviour, present is used verbatim, and no line derives it. `_atomic_write_text` exists **twice** and is byte-gated against its twin, so the rejection is mirrored — changing one copy would redden the gate and leave the hook-side path unprotected.

**The catch-up sweep reported inability as emptiness.** Four exits returned a bare `[]` — extensions unavailable, table missing, query raised, genuinely empty — while its one caller treats empty as nothing-to-do. Three of the four mean the opposite.

**The archive verdict now reads the sync outcome it already received.** The existing check counts occurrences in the target file only, so a projection landing in a different `CLAUDE.md` leaves that count at one and the archive reads clean. Only `wrote` claims a stray copy exists; the others say the outcome is unknown, because a failed sync can raise *after* the atomic rename.

## Three false claims removed

- A count in a string loaded into **every agent's context**, swept across all 13 sites. The true clause naming where the durable copy lives is kept.
- An over-broad rationale about which agents share a memory index.
- **A false equivalence claim in the merge guard's own comment.** It asserted that narrowing the flag-value arm left the matched language unchanged, "verified across 14 forms". Refuted by three independent measurements. The two forms stop the flag run in different places, so the predecessor can read a flag's numeric argument as the pull-request number — **an approval would bind to a different pull request than the one that merges.**

  The claim mattered because it was a **licence**: a maintainer reading "identical language" had written permission to broaden the arm back. Ordinary flag stacking does not reach it — 3772 real flag sets with one positional agree — but add a numeric flag value and all 168 enumerated forms diverge. A subject line that is a year; a body file named `1`.

  **The comment diagnoses this exact failure three paragraphs above its own commission of it.** The pattern itself is unchanged; the correction is comment-only, proven by a code-only diff rather than a suite pass.

## Guards added

- **Collection parity.** CI silently dropped 48 tests living inside the directories they guard. The invocation was already fixed; nothing stopped it re-narrowing. Now collects twice and compares **sets of node ids**, never counts. Guards both doors — the command line, and a `testpaths` config entry that re-narrows without touching it.
- **The managed-comment mirror.** Three modules write the same two managed-section comments, in three different syntactic shapes. The only thing keeping them in step was a sentence asking a maintainer to change all three at once — a claim, not an enforcement, which is the distinction this branch exists to draw. The comparison runs on source text with implicit concatenation collapsed first, because no single AST query reaches all three shapes and a phrase split across adjacent literals is not a contiguous substring of the source. Proven by mutation both directions: altering either mirror reddens one cell, altering the source of truth reddens both.
- **A stderr differential on the refusal path.** The sync-result contract already drove that path in a real subprocess and never asserted on stderr. Its two-outcome test now compares the arms' stderr as well as their disk evidence — differential rather than an emptiness check, because several pre-existing emitters can write to stderr when the sqlite and vector dependencies are absent and fire identically in both arms, so comparing the arms cancels them and measures only what the refusal adds.
- **The mint counter, which could not see a mint that created nothing.** A certification helper counted new tokens by filename, while the mint path retires the prior token by renaming it and then reuses the vacated name. So a mint producing nothing still counted 1 whenever a prior token existed to retire — and `minted == 1` is what proves a DENY is a *read* decision rather than a mint-side miss. The blind spot therefore sat on the detector for the recorded #1129 failure, where a faithful click minted zero tokens while every read-side unit stayed green: an over-block certified as a legitimate refusal. Counting now keys on **inode**, which a rename preserves and a fresh `O_EXCL` create does not. Filtering cannot fix it at any setting, because the defect is the identity rather than the pattern.
- **A pin on that counter**, so reverting it to a name diff goes red. It asserts the *hazard* was present rather than that the instrument ran — a retirement artifact on disk and zero live tokens surviving — because a zero-mint test that measures nothing passes exactly like one that works.
- **A consumer for the catch-up sweep's reason.** The sweep already distinguished four outcomes that used to be a bare `[]`; nothing read the distinction, and the function above it fell through to a clean `ok` while the caller above *that* discarded the dict entirely. A discarded return is syntactically invisible, which is why it survived review. The unanswerable cases now reach `get_status()` as `degraded`, a fault as `error`, and the specific reason travels alongside the coarse status.
- **A drift check on the zero-mint stub**, replacing the comment that asked a maintainer to keep it matching. It compares the observable post-state of the real function under a forced failure against the stub's, rather than a signature — a signature holds while the cleanup-versus-create order changes, and that order is the claim.
- **Capture pins** on the shape that discriminates a mis-partition. Three pre-existing pins use *alphabetic* dash-initial values and cannot witness it; they are now labelled for the property they actually pin.
- **A measured, ratified risk record** for an accepted low-severity finding, attributed rather than orphaned.

## Known residuals — stated, not closed

- **The anchor cannot validate its own choice.** The predicate answers *target-under-anchor*; it cannot answer *anchor-is-the-right-anchor*, because the anchor is one of its inputs. A caller declaring a directory it did not mean is honoured. Transport independence is not provenance independence.
- **Scope.** The refusal this exempts fires **only under a test process**. This closes the test-process ambient-write hazard, not a production one.
- **A nondeterministic certification test** (#1395), with a probable second instance recorded.
- **A pre-existing flake in a perf test this branch touched** (#1404), now measured rather than estimated. The suspicion was that doubling the workload against fixed wall-clock ceilings had eaten the margin. Measurement refutes it: the honest path is linear and the defect quadratic, so the doubling cost headroom above (13.55x → 6.82x, still comfortable) and *bought* 4.1x more discrimination below (3.40x → 13.85x). The real constraint is the **ratio** arm at 1.46x — and it is 1.47x at the merge base, so it is unchanged and not ours. It is also structurally bounded: honest 2.0 and defect 4.0 cap the total margin at 2.00x, so relaxing that ceiling spends the discriminator. Nothing to fix here.

## Verification

Full suite on a clean tree at HEAD: **13,857 passed, 15 skipped, 0 failed, 0 errors** — failure set verified empty by explicit scan, not read off the summary, with the tree state asserted in the same invocation before and after the run.

Guards proved by mutation in both directions, not by a green run. One earlier gate was **discarded as invalid** rather than reported: it ran while a file the suite imports was being edited, and a test resolving source text by cached line numbers read the wrong region. Not a flake and not a defect.

**Coverage, stated before the result rather than after it.** CI is one workflow of five steps; the local suite covers one of them. Two steps have no local analogue (a hosted checkout and a pinned interpreter setup). One is not covered because its tool is absent from this environment, re-verified rather than assumed. One dependency install is not reproduced. **CI is the gate; the local run is a pre-check.** An AST check found zero unused imports, which is not a claim that the lint step passes.

The merge-guard flag-value arm changed earlier in this branch's life, outside every reviewer's stated range. It is nevertheless covered — by committed tests that substitute the predecessor arm into the real pattern, monkeypatch the real module and drive the real extractor behind a non-vacuity guard, plus vendored baseline modules consumed by eight certification files comparing guard-level decisions. That certification re-runs, which no reviewer's prose does.

Version 4.6.27, PATCH: every change is additive with a default.


michael-wojcik:
## Fold: pact-memory embedding integrity, sync outcome channel, and a declared containment anchor

Three defects in one family — a signal true about a narrower proposition than the one being read for.

### What landed

**A save that stored no vector reported unqualified success.** The verification read back the memory row only, which cannot detect a missing vector by construction. The store path now reports its own capability.

**A refused sync and a suppressed sync were the same observation.** `sync_to_claude_md` returned a bare bool, so a write refused by the pytest guard and one suppressed by `--no-sync` both presented as "the file is unchanged". A negative control asserting the file changed could not tell them apart, and its sibling assertion went vacuous without going red. Now returns a result carrying `wrote / refused / suppressed / unresolved / missing / failed`, with `__bool__` preserving the previous meaning exactly so no caller changes behaviour. Carried across the process boundary into the CLI envelope.

**The containment check could not validate its own anchor.** Both callers derived the anchor from the same resolution that produced the target, so it asked whether a directory contained itself. A caller may now declare the anchor; absent means today’s behaviour, present is used verbatim, and no line derives it.

### Known residual: the anchor cannot validate its own choice

The refusal this parameter exempts fires only under a test process. In ordinary use the pytest signal is absent, the guard returns early, and nothing is refused — so there is nothing for the parameter to unblock there. **The parameter exists to make the guard’s exemption sound for test processes**, which is where the ambient-write hazard actually occurs: a suite writing into the operator’s real `CLAUDE.md`. The declared-anchor form replaces a blanket refusal with a containment guarantee, which is why the suppression arm regains its ability to discriminate.

**What the containment predicate can and cannot answer.** `_atomic_write_text` opens the target’s parent following symlinks and walks inodes up to the directory it was given as the anchor. That answers **target-under-anchor**. It cannot answer **anchor-is-the-right-anchor**, because the anchor is one of its inputs. So a caller that declares a directory it did not mean is honoured: the resolver’s first branch probes the declared directory, finds a `CLAUDE.md`, and the write lands beneath the declared root — contained, permitted, and wrong. This is a confused deputy. It is a property of the caller’s choice rather than of the mechanism, and no containment predicate can detect it.

**What the declared anchor does bound.** When the declared directory holds no `CLAUDE.md`, the resolver’s first branch misses and resolution continues to the worktree root and then the main repository root. Both are ancestors of a nested declared directory, and an ancestor never walks up into a descendant, so containment refuses. The remaining branch re-probes the locations the first branch already missed and cannot succeed where that failed. Measured against the live resolver, a declared directory nested inside a repository whose target resolves to that repository’s `CLAUDE.md` yields REFUSE, while an anchor re-derived from the resolution yields PERMIT.

**What the regression arm is, and is not.** It exercises the parameter’s plumbing and the containment verdict **through the test wiring**. It is **not** a replay of a production configuration: the archive path’s save carries an explicit suppression flag and its other leg is a read, so that path does not reach the sync at all. A test pins both the argv of each leg and the number of legs, so a new leg or a dropped flag fails it.

**Where the parameter is, and is not, available.** The paths that do sync in ordinary use are a plain command-line save without the suppression flag, and library callers of the save and search APIs, which sync by default. Neither has a wrapper that supplies the anchor. The parameter is therefore available on the API surface but unsupplied on the paths that actually reach the sync outside a test process; a library caller that wants the containment guarantee must pass it itself. This is a stated limit of the current shape rather than a defect in it — the guard those callers would be exempting does not fire for them in the first place, so the anchor buys them containment they can opt into, not a refusal they must escape.

**What this design does not claim.** Carrying the anchor as a function argument rather than through the environment makes it independent of the *resolver*. It does not make it independent of the *caller*. Transport independence is not provenance independence, and that is what leaves the deputy standing: a caller holding a single notion of the project spends that one value on both the boundary it declares and the inputs that steer resolution.

**Unchanged and unguarded by design:** an in-process caller syncing ambiently is exempt, because the suite’s own working-memory tests rely on that path.

### Alternatives searched and declined

- **A capability minted by the caller.** Vacuous where it would matter: minted from the same value used to steer the resolver, it designates the directory resolution was already going to reach. It addresses forgery of the anchor, a different problem from the anchor and the target sharing a source.
- **An inherited directory descriptor.** The resolver cannot fabricate one, so it satisfies unforgeability. Declined on practicality — descriptors are close-on-exec by default, so it must be defeated deliberately to work, and it does not port.
- **A creation-time predicate on the declared directory.** Declined because the field it needs is not exposed by `os.stat_result` on Linux, so it has no input where it would run and must fail closed, refusing every legitimate sandbox. An earlier objection argued a fresh checkout is younger than a sandbox; that tests absolute age rather than creation relative to process start and does not hold. The portability gap is the reason.

### Verification

Full suite green: **13754 passed, 15 skipped, 0 failed, 0 errors**, import hygiene pass.

Independent verification of the anchor commit confirmed all four priorities **by mutation with pre-registered predictions**, not by a green run:

- Breaking the permit direction reddens the positive control while the refusal arm still passes — demonstrating that a broken harness reaching a failure verdict would otherwise sail through.
- Stripping the rejection from one copy of the duplicated control reddens three tests including the byte-drift gate.
- Both new pins fire under their respective mutations.

`_atomic_write_text` exists twice and is byte-gated against its twin, so the anchor rejection is mirrored; a one-copy edit reddens the gate and leaves the hook-side path unprotected.

### Follow-ups filed

- **#1395** — a merge-guard certification test is nondeterministic (one tree, 27 failures then 26, byte-identical). A cert passing 39 times in 40 certifies less than it claims. Gate readings on this suite should carry the failure **set**, not a count.
- **#1396** — agent craft rules are invisible across stores (168 directories, 146 of them one-off spawn names).
- **#1392** — nothing detects or repairs an already-stale vector. Narrowed on the create side by this work, **unchanged on the detect side**.

